PR #21, opened 2025-06-19
Make R3D_SetRenderTarget const friendly

Caduckic:
Changed  R3D_SetRenderTarget to take const RenderTexture*. This allows calling it from const methods, example "Engine::Draw() const {}".


Bigfoot71:
Indeed, that makes more sense, thank you!

PR #24, opened 2025-07-01
Added bloom thresholds

grunthon:
Adds prefiltering by brightness for the bloom effect. I needed this for my own testing of the library so figured I'd share it.


Bigfoot71:
Great suggestion, thank you very much!

PR #25, opened 2025-07-01
Cleaner initialization of bloom prefilter

grunthon:


Bigfoot71:
Why not, but I actually wondered if these values could be precomputed as soon as `bloomThreshold` or `bloomSoftThreshold` are modified, rather than recalculated every frame?

That should be doable by keeping `bloomSoftThreshold` globally and storing `bloomPrefilter` instead of just `bloomThreshold`. Let me know what you think.

grunthon:
Good point and I had considered that. I'll look at caching it and only updating inside the Set functions when called. Feel free to close this until then.

PR #27, opened 2025-07-02
Cache bloom prefilter data

grunthon:
Let me know if the function to calculate the data is alright where it is.

Also technically storing bloomThreshold as a separate value is redundant now since it's stored inside R3D.env.bloomPrefilter.x for the shader uniform as well, but I think it's more readable and maintainable the way it is. If you'd prefer to not have it there, I can alter it.


Bigfoot71:
Yes, it's perfectly nice! And yes, I agree to keep `bloomThreshold`, it's perfect for me, thanks for the review!

PR #30, opened 2025-07-05
Fix for ternary operator with MSVC

grunthon:
MSVC doesn't like this ternary shorthand.


Bigfoot71:
Haha, I was just about to do it next! Thanks!

PR #32, opened 2025-07-06
Material emission assignment

grunthon:
I wanted to get the emission strength from Assimp to use for the multiplier using `AI_MATKEY_EMISSIVE_INTENSITY`, but it seems Assimp doesn't actually do anything with this. And for glb/gltf Assimp, it doesn't support the extentions for it assimp/assimp#6162.


Bigfoot71:
Oh, and I had thought of it but forgot, thanks!

PR #34, opened 2025-07-07
Dithering for debanding

grunthon:
As discussed in the issue, this really only works with at least a 16 bit floating point buffer.


Bigfoot71:
Super, thanks! I will take care of the necessary changes

PR #37, opened 2025-07-09
Fix comparison of material's albedo color alpha

grunthon:
Unsigned char `albedo.color.a` was being checked as if it were a float.


Bigfoot71:
A stupid mistake, thanks a lot!

PR #40, opened 2025-07-10
fixing R3D_GenMeshHeightmap

JettMonstersGoBoom:
account for Y in size passed into function.
fixes colors to WHITE and in correct range 0.0-1.0


Bigfoot71:
Thanks for the fix!

PR #41, opened 2025-07-11
unit scaling for loaded models

grunthon:
The assumed units and scaling isn't something that should be happening when importing models. This adds functions for having an optional scaling factor and I added it in the examples so they still work as they did.

I don't think any of this is really needed as it's up to the user to know and work with whatever units/scale they are using. Not really the primary reason to avoid it, but this scaling also does make model sizes inconsistent between R3D and Raylib.


grunthon:
I have an error in a function definition, fixing.

Issue #48, opened 2025-07-20
[REQUEST] Make Github Actions artifacts available


refinism:
Hi @Bigfoot71 , If you don't mind, could you preserve the artifacts from the GitHub actions, since the actions itself already compiled the tested build, I just hoping to use the shared/static library to avoid rebuilding every time, thanks 🤗

Bigfoot71:
Hi, I’m not against the idea, but I’m not very familiar with artifact retention. I’d just like to avoid any extra costs, maybe you know more about it than I do, but we could agree on a retention days value

~Also, if you’d like a bit more stability, you could always use pre-releases for now~

I wrote too quickly, it’s true that the binaries aren’t actually distributed with the pre-releases, sorry about that...

refinism:
I'll study the effects on the cost first, then I'll update the result later. (my current opinion: I don't think it'll affect the action cost other than the storage capacity)

Bigfoot71:
I just added artifact retention, they are kept for 14 days.
Note that builds will also be available permanently starting from the first stable releases.

PR #49, opened 2025-07-20
Update cmake-multi-platform.yml to upload artifacts

refinism:
#48 
Currently the archive isn't correct, don't merge


Bigfoot71:
Artifact retention has been added to the workflow

PR #51, opened 2025-07-23
fixed leak during R3D_Init

aalmkainzi:
Fixed a leak when loading blue noise.


Bigfoot71:
Oh, thank you for noticing that!

PR #54, opened 2025-07-24
recursive hierarchical model loading.

aalmkainzi:
I had a problem when loading 3D models with node tree when using `R3D_LoadModel`. (e.g. https://sketchfab.com/3d-models/mercedes-w13-concept-f1-wwwvecarzcom-1a1f8d992a3f4e2185603c87a0087781)

I tried loading the same model with raylib's `LoadModel` and it worked.

the new function I added recurses through the tree, and applies the correct transformation relative to its parent's transform.

I'm not sure if this is the best way to do it (I'm pretty new to understanding 3D models formats), but it correctly loaded the model that failed eariler.

I tested it with the r3d_animation example and that looked correct (had a problem with that earlier, turns out im not supposed to apply transformations if the mesh has bones).

this solves the problem I had, though I'm not sure if there's an edge case I'm not handling properly. There's probably a better solution.

### DIFF
--- a/src/r3d_model.c
+++ b/src/r3d_model.c
@@ -1963,7 +1963,7 @@ static inline Matrix r3d_matrix_from_ai_matrix(const struct aiMatrix4x4* aiMat)
 
 /* === Assimp Mesh Processing === */
 
-static bool r3d_process_assimp_mesh(R3D_Model* model, int meshIndex, const struct aiMesh* aiMesh, const struct aiScene* scene, bool upload)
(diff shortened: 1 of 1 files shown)


Bigfoot71:
I just tested it and yes, it even explains some of the scaling issues we were having, especially in the sponza scene!
I rushed the Assimp integration because I was short on time, so thank you very much for the review!

And I think it’s perfectly fine for the majority of cases imo

The only thing I noticed is that the normal matrix was being recalculated for each vertex in `r3d_process_assimp_mesh`, but I’ll take care of reviewing that, thanks again!

PR #55, opened 2025-07-26
diffuse/specular control for IBL

grunthon:
There is more I plan to do on this, like properly disabling IBL if there is zero intensity or possibly having a flag to do so, and adding controls for the visible skybox(like draw/don't draw, intensity, etc). I'm putting this up as it is to see if the approach seems alright. Mainly as far the implementation and separate diffuse/specular control.

As always naming things is the hard part and I'm not sure about all of it. IBL isn't really something that's been previously exposed to the user and has just been part of the skybox, so there's a bit of ambiguity.

### DIFF
--- a/src/r3d_core.c
+++ b/src/r3d_core.c
@@ -105,8 +105,9 @@ void R3D_Init(int resWidth, int resHeight, unsigned int flags)
     R3D.env.ambientColor = (Vector3) { 0.2f, 0.2f, 0.2f };
     R3D.env.quatSky = QuaternionIdentity();
     R3D.env.useSky = false;
-    R3D.env.iblDiffuse = 1.0f;
-    R3D.env.iblSpecular = 1.0f;
+    R3D.env.skyBackgroundIntensity = 1.0f;
+    R3D.env.skyAmbientIntensity = 1.0f;
+    R3D.env.skyReflectIntensity = 1.0f;
     R3D.env.ssaoEnabled = false;
     R3D.env.ssaoRadius = 0.5f;
     R3D.env.ssaoBias = 0.025f;
@@ -1345,8 +1346,8 @@ void r3d_pass_deferred_ambient(void)
                 r3d_shader_set_mat4(screen.ambientIbl, uMatInvProj, R3D.state.transform.invProj);
                 r3d_shader_set_mat4(screen.ambientIbl, uMatInvView, R3D.state.transform.invView);
                 r3d_shader_set_vec4(screen.ambientIbl, uQuatSkybox, R3D.env.quatSky);
-                r3d_shader_set_float(screen.ambientIbl, uIblDiffuse, R3D.env.iblDiffuse);
-                r3d_shader_set_float(screen.ambientIbl, uIblSpecular, R3D.env.iblSpecular);
+                r3d_shader_set_float(screen.ambientIbl, uSkyboxAmbientIntensity, R3D.env.skyAmbientIntensity);
+                r3d_shader_set_float(screen.ambientIbl, uSkyboxReflectIntensity, R3D.env.skyReflectIntensity);
 
                 r3d_primitive_bind_and_draw_screen();
 
@@ -1553,7 +1554,7 @@ void r3d_pass_scene_background(void)
     {
         glViewport(0, 0, R3D.state.resolution.width, R3D.state.resolution.height);
 
-        if (R3D.env.useSky)
+        if (R3D.env.useSky && R3D.env.skyBackgroundIntensity > 0.0f)
         {
             // Setup projection matrix
             rlMatrixMode(RL_PROJECTION);
@@ -1580,6 +1581,7 @@ void r3d_pass_scene_background(void)
 
                 r3d_shader_bind_samplerCube(raster.skybox, uCubeSky, R3D.env.sky.cubemap.id);
                 r3d_shader_set_vec4(raster.skybox, uRotation, R3D.env.quatSky);
+                r3d_shader_set_float(raster.skybox, uSkyIntensity, R3D.env.skyBackgroundIntensity);
                 r3d_shader_set_mat4(raster.skybox, uMatView, matView);
                 r3d_shader_set_mat4(raster.skybox, uMatProj, matProj);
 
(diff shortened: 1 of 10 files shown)


Bigfoot71:
It’s an interesting idea for artistic control, but I have some doubts

The thing is, I probably made it too simplified in the code, I wanted it to stay understandable for everyone, but I think it’s too dumbed down...

Actually, the "specular" here is just the direct reflection of the skybox, and the part I called "diffuse" is a bit more complicated because `Ks` is also part of the specular reflection

So the logical way would be `Ks *= iblSpecular` and `Kd *= iblDiffuse`, but I’m not sure if that level of control is really useful for anyone

And for the direct skybox reflection, tweaking it like this might feel weird, but honestly it could be interesting artistically in some cases

So here it’s better to think of "diffuse" as ambient light, and "specular" as the direct skybox reflection

If we keep these parameters as they are, it would probably make more sense to have something like:
```c
void R3D_SetSkyboxIntensity(float ambient, float reflection);
void R3D_GetSkyboxIntensity(float* ambient, float* reflection);
```

I’ll do some tests with the example skybox and your fork to see how it looks when playing around with this.

And sorry if the shader comments and variable names made it more confusing, I’ll think about how to make that clearer for everyone

___

Edit: I wrote the example function above with two parameters just to write it quickly, it’s not really a suggested signature, unless you like it x)

grunthon:
No problem. As I said I wasn't sure about it. I did wonder why IBL was part of ambient light and thought maybe it was so AO could directly affect it. I'll wait to hear if you come back with more thoughts, and don't be shy to let me know if it's something you'd rather handle on your own.

Also, just to point it out, specular is by definition a direct reflection. ;)

_edit: I do have a test/example I made for my testing if you're interested in it._
ss-ibl

Bigfoot71:
It’s fine for me! I replied a bit too fast but I took the time to check and test, looks perfect!

It's just that usually the sky reflection is just toggleable, but honestly it’s fine like this!

I was just wondering if it’d make more sense to put these params in `R3D_Skybox`?
Maybe it’s nicer to tweak it per skybox instead of global state?

Bigfoot71:
And I just meant maybe "ambient" and "reflection" are clearer names, they match what Unity and Godot use too
Also I mixed things up with the "non-IBL" mode, with IBL kS isn’t reused anyway, so forget that it's perfect!

_Sorry if it’s confusing, I kinda cheat and use a translator because I’m scared I won’t make sense, but it sometimes makes it worse x')_

___

Oh, and I’m distracted, everything’s fine, if you want to keep going like this I’ll end up merging it anyway!

About the two adjustable parameters, it’s fine, maybe "ambient" and "reflection" would be clearer, and maybe making it configurable per skybox would be better, if you disagree, feel free to say!

grunthon:
I was just kidding around about the term specular. It all gets messy with everyone using terms in different ways. Don't worry, you're perfectly clear. Using ambient and reflection seems good to me.

I do wonder if it would be best if IBL was separate from skyboxes, with the ability to use a skybox for IBL. They can serve different purposes and sometimes work best with different image formats. Right now even the included examples use non-HDR images for IBL which doesn't work very well. Plus having IBL as a separate system would make sense for features such as independent light and reflection probes. I'm not suggesting making those kind of changes now, or with this, just something to think about.

I'll work on this a bit more including the ideas you've brought up and post back sometime soon.

Bigfoot71:
Exactly, about the reflection probes, It’s something I’d like to add too, so I completely agree

I don’t really see how to unify all this right now, my first instinct would be to say that each object has its own parameters but they would all feed into the same uniforms in the end, but well…

Feel free to ping me, if I have an idea I’ll share it. Thanks for the thoughts!

_And yes, I’m aware about the skyboxes in the examples, even ignoring the HDR, they’re pretty dirty, I just try not to bloat the repo too much, I’m thinking about  slow connections, like mine x)_

grunthon:
I did a version of this where the intensity values are stored in `R3D_Skybox` so everything is per skybox. The problem is that since `R3D_EnableSkybox` makes a copy of the skybox into the environment data, the passed skybox object is not what is actually being used. That means it's back to the original situation that only values being used are essentially global ones just for the environment's copy of the skybox and storing the data in `R3D_Skybox` basically pointless. Hopefully I'm explaining that clearly.

The straightforward solution is to pass a pointer into`R3D_EnableSkybox` instead and store that in the environment data. Of course that leaves it to the user to keep the active `R3D_Skybox` alive and valid, and the problems that can come with that.

Let me know what you think about this, and apologies for this dragging out. I started this thinking it would be just scaling a few values.

Bigfoot71:
Oh yes, good point

> The straightforward solution is to pass a pointer into`R3D_EnableSkybox` instead and store that in the environment data. Of course that leaves it to the user to keep the active `R3D_Skybox` alive and valid, and the problems that can come with that.

The issue is that raylib aims to be accessible for developers of all levels, so R3D should be too
If users pass a reference to a skybox on the stack, we might get complaints about this later

Ideally, in this context, we should return the skybox by pointer instead of by copy, to avoid errors

So either we add more get/set functions, which is redundant and not very flexible, or we accept some small allocations, realistically, a program won’t recreate hundreds of skyboxes all the time

But maybe there’s a better solution too

Bigfoot71:
Also, I remember people telling me I worry too much about beginners, maybe overprotecting against possible mistakes isn’t good either x)

If you want, I can also run some "pattern tests" this week when I have time

grunthon:
You can look at the new commit and see what you think. What I decided on was storing the intensity values in the environment data since that's where the active skybox is anyway. It works out alright and makes sense without doing more substantial changes. I also changed to one function each for Get/Set. I considered just not having the Get function since it doesn't seem that useful and the user would probably be storing that data themselves if they needed it.

Bigfoot71:
I like it, if it’s ready to be merged, I’m fine with that

And I agree about the getters, honestly I’ve been thinking about removing them for a while too, since they’re redundant and not really useful. I originally added them to stay consistent with the lights, where they can actually be more useful.

With all this, I’m thinking about revisiting this whole environment system. Maybe we could keep all the setters but also add a description struct to configure everything at once, and with that, prepare the ground for an IBL system that’s less tightly coupled to the skybox

But for now, it’s perfect! If it’s good for you too I can merge

grunthon:
Yeah I think it's alright to merge if you do. It works fine and it doesn't seem worth spending more time on until bigger changes to IBL, as you say.

_If you were wondering, the small new commit was just to remove a change in the description for `R3D_Skybox` from when I was storing the intensity data there._

src/r3d_core.c
```diff
@@ -1549,7 +1554,7 @@ void r3d_pass_scene_background(void)
     {
         glViewport(0, 0, R3D.state.resolution.width, R3D.state.resolution.height);
 
-        if (R3D.env.useSky)
+        if (R3D.env.useSky && R3D.env.skyBackgroundIntensity > 0.0f)
```
Bigfoot71, line 1557: just a personal note to remind myself to check the conditions here

Bigfoot71:
Thanks for the PR, I took the time to read through everything just in case, I’m merging it!

PR #58, opened 2025-08-04
separate flags for skybox drawing and IBL

grunthon:
I didn't really like how I had the check for rendering the skybox so I split `R3D.env.useSky` into separate flags for drawing and IBL. I think it makes all the checks clearer and adds a little more distinction between skyboxes and IBL. Just let me know if you'd rather not have it this way.


Bigfoot71:
I find that having `drawSky` as well, which is just `skyBackgroundIntensity > 0.0f`, makes the logic slightly less straightforward, but since all of this will be revisited when we decouple the IBL from the skybox, I don't really mind either way

Personally, I wouldn't overthink it, but if this PR looks good to you and you prefer it this way, I'm fine with merging

grunthon:
That's fine. It's really only something I did in passing while working on something else. I'll just close this.

PR #61, opened 2025-08-17
Add depth of field (DoF) post-process effect

jensroth-git:
Introduces a depth of field (DoF) post-processing effect with new shader, API functions, and environment state. Adds example usage in examples/dof.c, updates CMake and shader loading logic, and exposes controls for focus point, scale, blur size, and debug visualization.

https://github.com/user-attachments/assets/b5d7ff73-5df9-41ff-9f13-e2b521f1443b


Bigfoot71:
Wow, great contribution! :D

I was actually hesitating recently about adding the DOF I made for a game I’m working on, since it can always be implemented on top of R3D without much trouble, but why not!

Just two points, the debug mode can be removed, and the texel size can be added as a uniform

I can merge it and take care of the doc and remaining tweaks if you want

jensroth-git:
Hey, Thanks, I would love to contribute some more in the future, 
we already briefly talked about making the material system scriptable on discord a while back, 
I think that could be awesome as well!

I have dreams of turning this into a code first game engine, 
theres way to few that havent devolved into UI / visual scripting nightmares :D 

oh and the debug modes were meant for dialing in the effect, similar to other game engines like UE5, 
I actually simply stole their visualization idea lol

image

image

but we could remove the linear depth one (mode 2) if you'd like.
~~iirc I based the texel size handling off of how its done in the other effects,~~ 
just checked and fxaa uses a uniform too, I dont know why I did it this way?
I guess we should just keep it consistent.

And thanks again for the great project!

Bigfoot71:
> I have dreams of turning this into a code first game engine,
> theres way to few that havent devolved into UI / visual scripting nightmares :D

Well, a rendering engine doesn’t make a full game engine, the workload of such a thing personally scares me x')
And yes, in any case, custom shaders are planned for soon, that will be a start!

Otherwise, yes, I understand the point of debug mode '1', it’s just that UE will reserve that only for its editor, the final shader will be compiled without any debug cases anyway, which is what bothers me a bit here

jensroth-git:
Ah, I see.
Maybe the python post-processing could handle stripping certain debug lines for shipping configurations? 
although I'm not a fan of premature optimizations.

for future work we should split the effect into multiple passes anyway,
but even as is, it’s already very usable, it's a very artist friendly DoF.

And I mean with raylib as the base, you cant really escape the Game Engine world anyway 😄

Bigfoot71:
Oh, it’s not even about premature optimization, having a debug mode in a production project feels a bit sketchy to me x)

Anyway, I’ll leave debug mode ‘1’ for now and fix it properly when I do the custom shaders. Like you said, the python scripts will get a revamp then too, so there’ll be a bunch of internal changes anyway

So I’ll just merge this and tweak it right now

For the shaders, I’ll just drop debug ‘2’ (we could add this visualization later, but not specifically in the DOF) and add uTexelSize as a uniform like the others. I’ll ping you here when it’s done if you want.

jensroth-git:
Thank you so much! 😄
looking forward to doing more contributions in the future!

would love to optimize this effect, and maybe turn it into a high quality bokeh DOF?
with some ideas from [here](https://catlikecoding.com/unity/tutorials/advanced-rendering/depth-of-field/) for example.

I think having a configurable bokeh shape could be awesome as well!
anyway, got lots of ideas, and motivation.

just let me know where you want to steer this!

Bigfoot71:
@jensroth-git I've done the review, mainly some cleanup and comments: https://github.com/Bigfoot71/r3d/commit/8970b5784e369f0be2e837b45d4d5d2d9c67eb31

For the shader, I removed debug mode '2' but kept mode '1', the function setter now only takes a bool and is briefly documented.

Regarding linearized depth, naming isn't very clear across versions, but the one you used in the DoF shader returns values between 'near' and 'far' (not normalized to 0-1), which is correct for world-space. I've added a note in the header comments.

Everything looks great, thanks! :D

Going forward, I'll focus on stabilizing existing features before adding new ones, the future goal is OpenGL ES support. Keep optimizations/features reasonable, no need for AAA quality, GL 3.3 and the "raylib-like" design have limits...

A configurable bokeh shape would be nice, that should be fine for this, but if some features require deep optimizations or too modern API features, we'll have to live with it :/

But if you have other ideas and want to implement them, it will always be very welcome!!

PR #64, opened 2025-08-20
Bilinary texture filter for smoothing.

GuvaCode:


Bigfoot71:
if you prefer why not

PR #66, opened 2025-08-22
adding support for custom ( user made ) animation systems

JettMonstersGoBoom:
adding support to allow custom animation systems. for possibly blending, ragdolls etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,8 +23,8 @@ endif()
 option(R3D_BUILD_DOCS "Build the doxygen doc" ${R3D_IS_MAIN})
 option(R3D_BUILD_EXAMPLES "Build the examples" ${R3D_IS_MAIN})
 
-option(R3D_RAYLIB_VENDORED "Use vendored raylib from submodule" OFF)
-option(R3D_ASSIMP_VENDORED "Use vendored assimp from submodule" OFF)
+option(R3D_RAYLIB_VENDORED "Use vendored raylib from submodule" ON)
+option(R3D_ASSIMP_VENDORED "Use vendored assimp from submodule" ON)
 
 # Set the cmake module path
 
--- a/examples/animation.c
+++ b/examples/animation.c
@@ -76,7 +76,7 @@ const char* Init(void)
 
     /* --- Load model animations --- */
 
-    anims = R3D_LoadModelAnimations(RESOURCES_PATH "dancer.glb", &animCount, 60);
+    anims = R3D_LoadModelAnimations(RESOURCES_PATH "dancer.glb", &animCount, 60, false);
 
     /* --- Setup scene lights with shadows --- */
 
--- a/include/r3d.h
+++ b/include/r3d.h
@@ -195,7 +195,6 @@ typedef enum R3D_Dof {
  * Controls wether to allow external animation matrices
  */
 
-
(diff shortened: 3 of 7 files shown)


JettMonstersGoBoom:
I've used an unnamed Union in R3D_ModelAnimation to avoid mostly duplicate structs, I don't mind switching back, this builds fine with GCC and Visual studio. let me know if this is an issue ?

Bigfoot71:
Mhhh, anonymous unions are C11, should we do it? Aaah x)

ps: _I’m going to read all of that tonight, in a few hours where I live_

JettMonstersGoBoom:
It's 2025, maybe it's time ? I'm fine rejigging this though.

Bigfoot71:
> It's 2025, maybe it's time ? I'm fine rejigging this though.

Yeah, it's actually good, I'll add a note in the readme at worst

However, why not have `R3D_LoadModelAnimations` take the `bool localPoses` parameter instead of duplicating the function with `R3D_LoadModelLocalAnimations`?

Especially since the version loading from memory is missing, that would create a double duplication...

If it's to keep the API as is, it's not really a problem imo as long as we're in pre-release. It's only from the first release onwards that we should avoid touching the functions (I should perhaps mention this now that I think about it)

Also, is using `Transform` rather than `Matrix` really necessary for local transformations?

I can merge and look at it myself later, it's just that I'm a little tired right now to get all the elements in mind x)

Just otherwise, 'skipRender' would be better declared at the end of the struct, otherwise, apart from the doubt about the transforms and the additional loading function, everything seems fine

JettMonstersGoBoom:
good call.
I was avoiding changing R3D_LoadModelAnimations as it's an API breaking change ? If you're ok with breaking it that's ok. a From Memory version is easier. 

My assumption and use case uses Transform instead of Matrix as they're easier to modify / interpolate. I'll look into that.

Bigfoot71:
> I was avoiding changing R3D_LoadModelAnimations as it's an API breaking change ?

Yes, clearly, but while we're in pre-release, we have to take advantage of it
I haven't been shy so far, and if you have any suggestions don't hesitate

> My assumption and use case uses Transform instead of Matrix as they're easier to modify / interpolate. I'll look into that.

Ah yes, clearly, rotations are not linear, so the interpolation will always be incorrect between start and end, forget it, we'll keep `Transform` and if I find a better solution than the union one day I'll let you know, but let's keep it like that for now then

There would be solutions, but it would just be an expensive mess

JettMonstersGoBoom:
added those changes and now have a conflict ! checking.. oh shadowcasting changed

Bigfoot71:
> added those changes and now have a conflict ! checking.. oh shadowcasting changed

Aaaah, I'm sorry, I forgot to say that. I had a suspicion that it might be a problem...
But it's fixed, thanks. Go merge!

Bigfoot71:
I've made a few adjustments here: https://github.com/Bigfoot71/r3d/commit/322857bb2b355a36965d9bb1b7d918550ad191a3

However, I'm thinking the naming in the union might not be clear x)
I don't know, if you have a better idea, I could modify it

Or if there is ever a change regarding the union I will keep you informed

JettMonstersGoBoom:
thanks for those. 
Yeah I'm thinking about too. ( I'll notice pretty quickly when it gets changed :) )

PR #69, opened 2025-08-23
move shadow cast mode out of materials

grunthon:
The idea here is that having per object shadow casting is generally more useful and common than per material. I initially tried including both but the added complexity didn't seem worth it and did not add any additional functionality. This moves the shadow casting mode to base objects(mesh/sprite) that is then used inside draw calls.

This does reorder the `R3D_ShadowCastMode` enum so that front face shadow casting is the default, which currently does apply to sprites as well which might be undesirable.

I think this is a better solution on the whole, but feedback is welcome. It does of course break compatibility with the previous method.

### DIFF
--- a/examples/sprite_instanced.c
+++ b/examples/sprite_instanced.c
@@ -39,7 +39,7 @@ const char* Init(void)
     texture = LoadTexture(RESOURCES_PATH "tree.png");
 
     sprite = R3D_LoadSprite(texture, 1, 1);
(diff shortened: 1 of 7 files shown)


Bigfoot71:
Yes, I confirm, it's much more flexible and "consistent" with what we see elsewhere

And no problem with the changes, as long as we're in pre-release, we should improve the API as much as possible!

PR #70, opened 2025-08-23
removed premature allocations

grunthon:
Draw calls were being allocated before the checks to see if they would even be used.

### DIFF
--- a/src/r3d_core.c
+++ b/src/r3d_core.c
@@ -441,11 +441,11 @@ void R3D_End(void)
 
 void R3D_DrawMesh(const R3D_Mesh* mesh, const R3D_Material* material, Matrix transform)
 {
-    r3d_drawcall_t drawCall = { 0 };
-
     if (mesh == NULL || mesh->skipRender) {
         return;
     }
+
+    r3d_drawcall_t drawCall = { 0 };
 
     switch (material->billboardMode) {
     case R3D_BILLBOARD_FRONT:
@@ -490,11 +490,11 @@ void R3D_DrawMeshInstancedPro(const R3D_Mesh* mesh, const R3D_Material* material
                               const Color* instanceColors, int colorsStride,
                               int instanceCount)
 {
-    r3d_drawcall_t drawCall = { 0 };
-
     if (mesh == NULL || mesh->skipRender || instanceCount == 0 || instanceTransforms == NULL) {
         return;
     }
+
+    r3d_drawcall_t drawCall = { 0 };
 
     drawCall.transform = globalTransform;
     drawCall.material = material ? *material : R3D_GetDefaultMaterial();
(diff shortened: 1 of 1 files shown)


Bigfoot71:
AFAIK, without optimizations in either case, most compilers will follow a predictable approach and should allocate at the beginning of the call, regardless of order

And with O2/O3, it shouldn't be stupid enough to allocate before early exit

I did the test on compiler explorer

For that:
```c
void f0(int x) {
    int y = 42;
    if (x == 0) return;
    printf("%d\n", y);
}

void f1(int x) {
    if (x == 0) return;
    int y = 42;
    printf("%d\n", y);
}
```

We get this with -O0 on GCC 13.3:
```asm
f0(int):
        push    rbp
        mov     rbp, rsp
        sub     rsp, 32    ; Alloc here
        mov     DWORD PTR [rbp-20], edi
        mov     DWORD PTR [rbp-4], 42
        cmp     DWORD PTR [rbp-20], 0
        je      .L4
        mov     eax, DWORD PTR [rbp-4]
        mov     esi, eax
        mov     edi, OFFSET FLAT:.LC0
        mov     eax, 0
        call    printf
        jmp     .L1
.L4:
        nop
.L1:
        leave
        ret

f1(int):
        push    rbp
        mov     rbp, rsp
        sub     rsp, 32    ; Alloc here
        mov     DWORD PTR [rbp-20], edi
        cmp     DWORD PTR [rbp-20], 0
        je      .L8
        mov     DWORD PTR [rbp-4], 42
        mov     eax, DWORD PTR [rbp-4]
        mov     esi, eax
        mov     edi, OFFSET FLAT:.LC0
        mov     eax, 0
        call    printf
        jmp     .L5
```

So it directly allocates 32 bytes on the stack from the beginning in both cases

And for O3, it gives this:
```asm
f0(int):
        test    edi, edi
        jne     .L4
        ret
.L4:
        mov     esi, 42    ; Stored in a register
        mov     edi, OFFSET FLAT:.LC0
        xor     eax, eax
        jmp     printf

f1(int):
        test    edi, edi
        jne     .L7
        ret
.L7:
        mov     esi, 42    ; Stored in a register
        mov     edi, OFFSET FLAT:.LC0
        xor     eax, eax
        jmp     printf
```

In this case, the two codes are strictly identical and 'y' is stored in a register after the early exit

I'm just sharing because it piqued my curiosity, we can merge because why not, but it seems to be a detail

Here is the link if you want to test or correct me: https://godbolt.org/z/Pae1hfs1s

grunthon:
I did suspect that might be the case but didn't check the compiler output. I do generally avoid trying to outsmart the compiler but for something straightforward like this I'll usually at least try to make it explicit in the source.

Interesting to see your results and thanks for taking the time to check.

PR #71, opened 2025-08-24
SSAO intensity and power controls

grunthon:
This works similar to Godot's controls for applying SSAO. I added some adjustments in the Sponza example that can be used(or removed if not wanted).

### DIFF
--- a/examples/sponza.c
+++ b/examples/sponza.c
@@ -24,6 +24,9 @@ const char* Init(void)
 
     R3D_SetSSAO(true);
     R3D_SetSSAORadius(4.0f);
+    R3D_SetSSAOIntensity(1.25f);
+    R3D_SetSSAOPower(1.25f);
+
     R3D_SetBloomMode(R3D_BLOOM_MIX);
 
     /* --- Set default ambient color (when no skybox is activated) --- */
--- a/include/r3d.h
+++ b/include/r3d.h
@@ -2230,6 +2230,49 @@ R3DAPI void R3D_SetSSAOIterations(int value);
  */
 R3DAPI int R3D_GetSSAOIterations(void);
 
+/**
+ * @brief Sets the intensity multiplier for the SSAO effect.
+ *
+ * This function sets the the base multiplier used by the SSAO effect.
+ * Higher values will result in darker occlusion.
+ *
+ * @param value The intensity multiplier for SSAO.
+ *
+ * Default: 1.0
+ */
+R3DAPI void R3D_SetSSAOIntensity(float value);
+
+/**
+ * @brief Gets the intensity multiplier for the SSAO effect.
+ *
+ * This function retrieves the intensity multiplier applied to the SSAO effect.
+ *
+ * @return The intensity multiplier for SSAO.
+ */
+R3DAPI float R3D_GetSSAOIntensity(void);
+
+/**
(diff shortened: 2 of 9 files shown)


Bigfoot71:
It's great! And for the example it's very good as well, I think anyway to revisit Sponza to make it fully configurable to have a good demo, so it will allow me to think about it, thanks!

grunthon:
Related, I'm almost finished adding the ability for occlusion to affect diffuse lighting. Right now I have it only from SSAO but it could easily also use occlusion from the ORM texture. I just hadn't decided on how it would be controlled if that was the case. Two separate settings for SSAO and the texture? Just one for both? Wondered if you have any opinion on it before I make a PR.

Bigfoot71:
Oh, I was planning to do that after reviewing the forward rendering, but great!

As for the "light affect" for the occlusion map, I think it should be in R3D_Material, in any case, Godot as a reference again, that’s how it’s done there

image

grunthon:
I'll just keep it as I have it, with a global light affect value and for SSAO only. The occlusion map being per material can be added later. I've not bothered trying to get any of this working in forward rendering since I know it needs work first.

Getting more off topic about forward rendering, I've been thinking about it as well (I'm actually not a big fan of deferred rendering). Off and on for a while now I've been experimenting with tiled/clustered shading for forward rendering. This does use compute shaders and SSBOs which of course means GL 4.3. That's a whole different topic and something I know has come up before. I might make a discussion about all this sometime if you're open to it.

Bigfoot71:
TL;DR I don’t think we’ll move to GL 4.3 or use more modern features in r3d, despite **many** regrets about it...

___

> Getting more off topic about forward rendering, I've been thinking about it as well (I'm actually not a big fan of deferred rendering). Off and on for a while now I've been experimenting with tiled/clustered shading for forward rendering. This does use compute shaders and SSBOs which of course means GL 4.3. That's a whole different topic and something I know has come up before. I might make a discussion about all this sometime if you're open to it.

Yeah, I’m not a big fan of deferred either. I went that route for r3d because it was the most flexible way to efficiently manage the number of lights, while staying compatible with GL 3.3 so anyone could use it with raylib without losing too much time.

Since then, a lot has changed, so some things could still be questioned, but GL 4.3 remains problematic for MacOS users. Some people seem to already be using r3d on Mac, so I can hardly say “no, get lost”...

However, in parallel, I’ve started working on an experimental project with a C API and a philosophy close to raylib, but fully 3D oriented (2D limited to UI if any), with also real scene and physics management, and I have forward+ in mind too.

It’s not public yet and still very experimental in terms of design, without any pretension of matching standard engines, and it’s not meant to replace r3d. The idea is to address my biggest regrets with r3d while avoiding some internal design choices I didn’t like.

Raylib is amazing for simple projects that look great, but I also want to push things further in the same style, and the raylib ecosystem becomes a hindrance for that, hence the need for a more focused library imo.

We’ll see if that turns out to be viable, if it pops up one day, it means it was meant to be...

All that to say, that for now, I think r3d should remain simple on its own, although of course that doesn’t stop us from do our best within its limitations

___

However, if you want to share criticism, objections, or even crazy solutions, I’d be glad to hear them!

**And if you want to start a discussion on the repo, you're free, it’s just a couple of clicks (not counting the text), I’ll never close it right away.**

PR #73, opened 2025-08-25
ssao light affect for direct lighting

grunthon:
You can see what you think. I don't particularly like doing it in the scene shader, but it seemed the most reasonable place to get the accumulated lighting in deferred rendering.

I know you're going to rework the Sponza scene, but I did make a few tweaks that you can revert if you want. The initial camera position was driving me crazy ;)

### DIFF
--- a/examples/sponza.c
+++ b/examples/sponza.c
@@ -25,13 +25,14 @@ const char* Init(void)
     R3D_SetSSAO(true);
     R3D_SetSSAORadius(4.0f);
     R3D_SetSSAOIntensity(1.25f);
-    R3D_SetSSAOPower(1.25f);
+    R3D_SetSSAOPower(1.5f);
 
     R3D_SetBloomMode(R3D_BLOOM_MIX);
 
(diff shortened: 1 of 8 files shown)


Bigfoot71:
No, that seems perfect, I don’t see how it could have been done any better
And yeah, the camera was actually set up randomly x)

GuvaCode:
> You can see what you think. I don't particularly like doing it in the scene shader, but it seemed the most reasonable place to get the accumulated lighting in deferred rendering.
> 
> I know you're going to rework the Sponza scene, but I did make a few tweaks that you can revert if you want. The initial camera position was driving me crazy ;)

Wow!!! It looks great!!!, TNX. But I'm already tired of bindings edit :) 
w

grunthon:
Ha. Perhaps only doing it for tagged releases might be good for your sanity.

GuvaCode:
> Ha. Perhaps only doing it for tagged releases might be good for your sanity.

The advice is good, but not for me. I'm used to changing bindings almost in real time. This has been the case with raylib for 5 years now.

PR #75, opened 2025-08-25
control of mipmap levels for bloom

grunthon:
Right now the maximum possible amount of mipmap levels are used for the bloom effect. This is what you'd want for a lot of other uses of mipmaps, but doesn't usually give good results for bloom/glow.

The issue is that with the maximum amount of mipmaps any part of the image can have a significant impact on the whole output since it's dispersed across the entire image. This can really wash out the result (like an early 2000s "bloom everywhere!" game).

Here's an example.

Maximum possible (in this case 10)
bloom-chain-max-10

Using a more reasonable amount (6)
bloom-chain-6

This add the ability to set the number of levels, but it does still allow the maximum to be used if that is desired. The default I chose is around a typical amount used elsewhere. I did expose `r3d_target_load_mip_chain_hs` in the r3d_state header so it can be accessed (I think it used to be like this?). There could also be a corresponding function to unload to avoid duplicating that code if this seems alright. Just let me know thoughts about this.

### DIFF
--- a/include/r3d.h
+++ b/include/r3d.h
@@ -2411,6 +2411,30 @@ R3DAPI void R3D_SetBloomMode(R3D_Bloom mode);
 R3DAPI R3D_Bloom R3D_GetBloomMode(void);
 
 /**
+ * @brief Sets the number of mipmap levels used for the bloom effect.
+ *
+ * This function controls how many mipmap level are generated for use in the bloom effect.
+ * More levels will give a smoother and more widely dispersed effect, while less mipmaps
+ * can provide a tighter effect. Setting this value to 0 will result in the maximum
+ * possible amount of levels to be used. Use of this function will rebuild the
+ * mipmaps and may give a one time performance hit.
+ *
+ * @param value The number of mipmap level to be used for the bloom effect.
+ *
+ * Default: 7
+ */
+R3DAPI void R3D_SetBloomLevels(int value);
+
+/**
+ * @brief Gets the current amount of mipmap levels used for the bloom effect.
+ *
+ * This function retrieves the current amount of mipmap levels in use by the bloom effect.
+ *
+ * @return The number of mipmap level currently used for the bloom effect.
+ */
+R3DAPI int R3D_GetBloomLevels(void);
+
+/**
  * @brief Sets the bloom intensity.
  *
  * This function controls the strength of the bloom effect. Higher values result
--- a/src/r3d_core.c
+++ b/src/r3d_core.c
@@ -123,6 +123,7 @@ void R3D_Init(int resWidth, int resHeight, unsigned int flags)
     R3D.env.ssaoPower = 1.0f;
     R3D.env.ssaoLightAffect = 0.0f;
     R3D.env.bloomMode = R3D_BLOOM_DISABLED;
+    R3D.env.bloomLevels = 7;
     R3D.env.bloomIntensity = 0.05f;
     R3D.env.bloomFilterRadius = 0;
     R3D.env.bloomThreshold = 0.0f;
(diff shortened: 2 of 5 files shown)


Bigfoot71:
Great, as always!

For the exposure of `r3d_target_load_mip_chain_hs` yes that’s very good
And yes, we could also add an unloading function as well

So nothing else to say, I guess I can merge directly

I’ll add the unloading function so we don’t waste any time

___

_but I kinda liked the vaseline effect in Oblivion x')_

Bigfoot71:
I added the unload function: https://github.com/Bigfoot71/r3d/commit/ac971489befa9873ccd709f7c8cbd1d3a05c4caa

However I would like to point this out, you were doing this in `R3D_SetBloomLevels`:
```c
void R3D_SetBloomLevels(int value)
{
	if (R3D.target.mipChainHs.chain != NULL) {
		for (int i = 0; i < R3D.target.mipChainHs.count; i++) {
			glDeleteTextures(1, &R3D.target.mipChainHs.chain[i].id);
		}
		RL_FREE(R3D.target.mipChainHs.chain);

	    r3d_target_load_mip_chain_hs(R3D.state.resolution.width, R3D.state.resolution.height, value);
	}

	// Update value based on actual number of mip levels generated
	R3D.env.bloomLevels = R3D.target.mipChainHs.count;
}
```

The problem is that if the mip chain is non-null and a higher number is specified for example, the mip chain is not re-generated

So I modified it to unload it only if non-null and always regenerate it afterwards:
```c
void R3D_SetBloomLevels(int value)
{
	if (R3D.target.mipChainHs.chain != NULL) {
		r3d_target_unload_mip_chain_hs();
	}

	// Re-generate the mip chain
	r3d_target_load_mip_chain_hs(R3D.state.resolution.width, R3D.state.resolution.height, value);

	// Update value based on actual number of mip levels generated
	R3D.env.bloomLevels = R3D.target.mipChainHs.count;
}
```

However, we could also review this by avoiding deleting everything and re-creating if the number indicated is lower than what is already present, for example

grunthon:
Ahh I should have noticed that. I was rearranging some things before I submitted and apparently I rearranged a little too much. Thanks for catching it.

I did think about altering the chain instead of rebuilding the whole thing but just stayed with keeping it simple. I'd hope nobody is trying to call it every frame or something.

grunthon:
Actually looking at it again I believe my reasoning was that if the chain was null, that means bloom hasn't been enabled so there's no reason to build the chain. That will happen in `r3d_framebuffer_load_bloom` via `R3D_SetBloomMode` when bloom is enabled. 

But then the problem would be that `R3D_SetBloomLevels` would not set `R3D.env.bloomLevels` to the desired value if bloom hasn't been enabled yet, which could seem odd. Hm.

Bigfoot71:
Mhh yes, so something clearer would be to do this:
```c
void R3D_SetBloomLevels(int value)
{
	if (R3D.env.bloomMode != R3D_BLOOM_DISABLED) {
		r3d_target_unload_mip_chain_hs();
		r3d_target_load_mip_chain_hs(R3D.state.resolution.width, R3D.state.resolution.height, value);
	}

	// Update value based on actual number of mip levels generated
	R3D.env.bloomLevels = R3D.target.mipChainHs.count;
}
```

The same as before, but more explicit in the idea, because we're not checking the mipchain but the bloom state

The mip chain should never be NULL here, and there's also an assert in `r3d_target_unload_mip_chain_hs()`

And yes, it's also better if you want to define the number of levels before actually activating the bloom, and in the other case it could pose a problem

Bigfoot71:
I just pushed the change, you will tell me if necessary, thank you for coming back to it

grunthon:
That's definitely a better way to handle it and easier to read.

PR #76, opened 2025-08-26
flag shadowmap for update on creation

grunthon:
Sets the `shouldUpdate` flag to true when creating a shadowmap. If the shadowmap for a light is modified, like changing the resolution with `R3D_EnableShadow`, and the light is not set for continuous update it's left in an incorrect state until the next update.

It's debatable if this is needed and the user should have to manually update when modifying a shadowmap since that's how they configured it, but if nothing else this corrects a typo for `shouldUpdate`.

### DIFF
--- a/src/details/r3d_light.c
+++ b/src/details/r3d_light.c
@@ -142,7 +142,7 @@ void r3d_light_init(r3d_light_t* light, R3D_LightType type)
     light->shadow.updateConf.mode = R3D_SHADOW_UPDATE_INTERVAL;
     light->shadow.updateConf.frequencySec = 0.016f;
     light->shadow.updateConf.timerSec = 0.0f;
-    light->shadow.updateConf.shoudlUpdate = true;
+    light->shadow.updateConf.shouldUpdate = true;
 
     /* --- Set specific shadow config --- */
 
@@ -178,6 +178,8 @@ void r3d_light_create_shadow_map(r3d_light_t* light, int resolution)
         light->shadow.map = r3d_light_create_shadow_map_omni(resolution);
         break;
     }
+
+    light->shadow.updateConf.shouldUpdate = true;
 }
 
 void r3d_light_destroy_shadow_map(r3d_light_t* light)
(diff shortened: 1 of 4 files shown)


Bigfoot71:
Mhhh, I admit that’s a point I hadn’t thought about, mainly because as soon as I finished making this ID-based system, I already didn’t really like it.

I started with the idea, for simplicity, that each light was immutable in type, and so I must have implicitly thought the same thing for shadows. Maybe some parts of the code partially handle this kind of case, but I forgot and I should have looked into it again, though this system is a bit of a blocker for me...

In any case, I clearly must not have thought that someone might want to resize the shadow maps by calling `R3D_EnableShadow` twice in a row.

Honestly, I don’t really know what to think of it...

If you have a solution that could prevent strange behavior or anything like that, I’d accept it. But if you also have ideas for approaches that would be better than the current one, or even just a small API redesign, I’m open to it, as long as we’re still in pre-release...

grunthon:
I can't say much about the general approach without thinking more about it. I had just run into this when integrating options for shadows in a project I'm working on.

Off the top of my head maybe having `R3D_EnableShadow` only take a bool for enabling/disabling shadows and then a separate function for setting the resolution might make more sense. At least for this specific issue.

Bigfoot71:
Mhhh, the problem is that we’d need to take into account the call that defines the resolution before creating the shadow map, and even in practice it’s not necessarily clear to just anyone right away

Well, I’ll think about it, I’ll probably make some changes soon and ping you here directly if you’d like

grunthon:
It seems reasonable to just have the resolution setting function return if there isn't a shadow map already assigned to the light, probably with a log message. `R3D_EnableShadow(true)` could generate a set default size map if one doesn't already exist. It already does that now if it's passed 0 for resolution. That would mean creating/destroying textures more than needed when doing `R3D_EnableShadow(true); SetShadowResolution(4096);` but it does keep things simple.

Another option is storing a resolution in `r3d_shadow_t` even when there isn't already a texture. That would avoid any confusion about having to enable/set resolution in any particular order.

Just some quick thoughts. If I come up with anything more clever before you find a solution, I'll let you know.

PR #78, opened 2025-08-27
Added Boneoverride example

JettMonstersGoBoom:
custom_animation.c shows 3 uses of animation
"internal" "world matrices", and "local hierarchy".
it's is expected that ALL 3 animations will visually appear identical.

also now skip animation update if custom matrices are used.


Bigfoot71:
Perfect, super cool, thanks! I’ll look into this scaling issue!

PR #80, opened 2025-08-28
additional options for shadow cast modes

grunthon:
I tired multiple approaches to this, but what I was trying to address was shadow casting modes and face culling having become somewhat inflexible. This allows any combination of enabling/disabling shadow casting and what face culling is used as well as using the object's material to determine face culling.

As I said, I tried different ways to do this (like separate options for casting mode and culling mode) and there still might be a better way to handle this. Open to any feedback, but I do think this functionality should exist.

### DIFF
--- a/examples/sprite_instanced.c
+++ b/examples/sprite_instanced.c
@@ -39,7 +39,7 @@ const char* Init(void)
     texture = LoadTexture(RESOURCES_PATH "tree.png");
 
     sprite = R3D_LoadSprite(texture, 1, 1);
-    sprite.shadowCastMode = R3D_SHADOW_CAST_DOUBLE_SIDED;
+    sprite.shadowCastMode = R3D_SHADOW_CAST_ON_DOUBLE_SIDED;
 
     /* --- Create multiple transforms for instanced sprites with random positions and scales --- */
 
--- a/include/r3d.h
+++ b/include/r3d.h
@@ -140,10 +140,15 @@ typedef enum R3D_CullMode {
  * and if so, whether it is also rendered in the main pass.
  */
 typedef enum R3D_ShadowCastMode {
-    R3D_SHADOW_CAST_ON,             ///< The object casts shadows; the faces used are determined by the mesh culling mode.
-    R3D_SHADOW_CAST_DOUBLE_SIDED,   ///< The object casts shadows with both front and back faces, ignoring face culling.
-    R3D_SHADOW_CAST_ONLY,           ///< The object does not render normally, but still contributes to shadow maps.
-    R3D_SHADOW_CAST_DISABLED        ///< The object does not cast shadows at all.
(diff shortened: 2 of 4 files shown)


Bigfoot71:
I really like it!

I had revisited this just to allow shadow cast only, and again, I had followed the same logic as Godot, which uses the object’s face culling when not double-sided.

But now, this is very flexible, nothing to complain about. Go for it!

___

And I don’t think we need to go any further, the number of enums might seem verbose, but I find it very explicit.
I don’t think we could do better while keeping the same flexibility, that’s perfectly acceptable imo

Bigfoot71:
Ah, we could have used "simple numbering" for the enum and done the shift in the function that checks shadow cast only

But well, both approaches have their pros and cons, just a thought

grunthon:
Yeah I got myself going in circles a bit trying to settle on the best method, and wasn't sure about so many values in the enum. It does work fine and avoids any overly complicated logic. Obviously feel free to make any changes.

grunthon:
Something that just occurred to me is how close the `R3D_SHADOW_CAST_ON` and `R3D_SHADOW_CAST_ONLY` modes are in spelling which could make mistakes easier. Maybe a different naming convention would be better. Hmm.

Bigfoot71:
Yeah, I removed the shifts from the enum, and added a macro for internal checks: https://github.com/Bigfoot71/r3d/commit/82d83f529715ac8303f286642e5651a60c061ecf

That’ll prevent giving the impression that they can be combined

Yeah, true, I don’t see how to better describe the ‘ONLY’ mode, and for ‘ON’, apart from longer alternatives like ‘ENABLED’, ‘ACTIVE’, ‘AUTO’, ‘DEFAULT’, I’m not really sure, they all feel just as "awkward" to be honest, but I agree

grunthon:
Maybe just `R3D_SHADOW_CAST_ON` and `R3D_SHADOW_ONLY`?

Bigfoot71:
> Maybe just `R3D_SHADOW_CAST_ON` and `R3D_SHADOW_ONLY`?

Meh, that also feels weird to me

Or we split the mode and the culling?

Something like a global directive:
```
R3D_SHADOW_CAST_ENABLED,
R3D_SHADOW_CAST_ONLY,
R3D_SHADOW_CAST_DISABLED
```

And the culling separately:
```
R3D_SHADOW_SIDE_AUTO,
R3D_SHADOW_SIDE_FRONT,
R3D_SHADOW_SIDE_BACK,
R3D_SHADOW_SIDE_BOTH
```

First is default

___

It seems so obvious, sorry if you thought of it before the PR x)
But yeah, I think it might be best in the end
And it solves this problematic naming

grunthon:
I did have it that way for a while. It just wound up being more logic in more places. At least for me. That doesn't mean it might not be better for the user. If you can do it cleaner, by all means. I might look over it again later, but I think I'm done with it for the moment ;)

PR #91, opened 2025-09-22
remove piping when processing shaders

grunthon:
As far as I know cmake commands don't support piping, and it was causing errors. There's likely a cleaner way to handle this than using a temporary file (a wrapper script that calls the other two, or possibly combining them) but this does fix the issue.

### DIFF
--- a/cmake/EmbedShaders.cmake
+++ b/cmake/EmbedShaders.cmake
@@ -8,9 +8,11 @@ function(process_shader shader_file output_file_var)
     add_custom_command(
         OUTPUT "${output_file}"
         COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/generated/include/shaders"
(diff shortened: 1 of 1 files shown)


Bigfoot71:
Great, I was planning to take care of that, I'm slow with r3d at the moment because I just released this: https://github.com/Bigfoot71/Hyperion/tree/master

This is the test of the "raylib/r3d all-in-one" library I mentioned in another issue or PR

I’ll add your fix to this project, and quite a few things done here will also make their way into r3d

Thanks for the fix!

___

And honestly, as a temporary solution, I think it's okay. I’ll deal with the script stuff later, it’s been driving me crazy x')

grunthon:
> This is the test of the "raylib/r3d all-in-one" library I mentioned in another issue or PR

Ah, I've been curious about that. I will definitely be taking a look at it.

PR #94, opened 2025-09-26
cmake doxygen check

grunthon:
CMake generation would fail if doxygen wasn't available and r3d was considered the main project, regardless of if the documentation was wanted. This changes the option to be manually set and gives an error if doxygen isn't found.

An alternative would be to leave the option as it was(default to on when r3d is the main project) and give a warning if doxygen isn't found, if that would be preferred.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ endif()
 
 # Project configuration options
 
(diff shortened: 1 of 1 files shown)


Bigfoot71:
No, I think it's fine like that, and yeah definitely it shouldn't have been the default, thanks!

PR #95, opened 2025-09-29
improve texture filtering

grunthon:
Raylib's `SetTextureFilter` is very specific with what it sets depending on the filtering mode used and does not set `GL_TEXTURE_MIN_FILTER` and `GL_TEXTURE_MAG_FILTER` when setting any anisotropic filtering mode. Calling `SetTextureFilter` with `TEXTURE_FILTER_TRILINEAR` will set the values the way you would expect, and then anisotropic filtering can be set after that.

It doesn't read very well and I've actually had code to handle this without using Raylib's functions since before I started using R3D. I might adapt that and replace this later, if that would be wanted.

Current result of R3D_SetTextureFilter(TEXTURE_FILTER_ANISOTROPIC_16X)
sa

And with this change
sb

### DIFF
--- a/src/r3d_model.c
+++ b/src/r3d_model.c
@@ -2353,6 +2353,23 @@ static Image r3d_load_assimp_image(
     return image;
 }
 
+static r3d_apply_texture_filtering(Texture2D* texture)
+{
+    if (R3D.state.loading.textureFilter > TEXTURE_FILTER_BILINEAR) {
+        GenTextureMipmaps(texture);
+
+        // NOTE: Needed to set min and mag filter even when using anisotropic filtering
+        SetTextureFilter(*texture, TEXTURE_FILTER_TRILINEAR);
+
+        if (R3D.state.loading.textureFilter > TEXTURE_FILTER_TRILINEAR) {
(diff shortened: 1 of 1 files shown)


Bigfoot71:
Ah perfect, it had to be managed effectively, thank you!

Bigfoot71:
> It doesn't read very well and I've actually had code to handle this without using Raylib's functions since before I started using R3D. I might adapt that and replace this later, if that would be wanted.

If you want to propose your method and it has any advantages I’m not against it
I’ll probably revisit this kind of thing soon as well, so it’s up to you

grunthon:
> If you want to propose your method and it has any advantages I’m not against it

Not really, at least not without adding other functionally I'm doing elsewhere. It's just weird in a way I don't like.

I can see the logic behind why `SetTextureFilter` works how it does but it's odd that getting ansiotropic filtering set how you would expect is
```
SetTextureFilter(texture, TEXTURE_FILTER_TRILINEAR);
SetTextureFilter(texture, TEXTURE_FILTER_ANISOTROPIC_16X);
```
At least with just the base Raylib functions. I thought about bringing this up on the Raylib repo when I first encountered it. Maybe I will.

PR #97, opened 2025-09-30
Pipeline review

Bigfoot71:
- Merged instanced and non-instanced shaders  
  -> everything now handled in the same shader with uniform conditions  

- Billboards fully computed on the GPU  
   -> removed CPU-side matrix management for billboarding  

- Reduced shader program switches  
   -> not only for instanced rendering but also for every draw call for shadow maps  

- Added alpha cutoff to deferred rendering  
   -> no clue why it wasn’t there before...

- Fixed instance alpha not being considered for alpha cutoff in shadow maps  

- Removed the depth pre-pass
   -> which was broken anyway and had become counterproductive/useless in the current state of things.

- Probably a few other tiny cleanups I did on the fly and forgot to note (oops)

___

~NOTE: Right now there’s a problem with instance transforms and also instance colors.
Haven’t dug into the transform issue yet. For colors, it’s just because the default attribute is literally black, sooo yeah...
We’ll need to set valid default attributes for instances in each VAO, bind them, then clean up on each instanced render.~

Also, wouldn’t it be nicer to manage instance buffers explicitly instead of creating/copying data into a temp VBO every draw call? That’s what raylib does too, but it’s SUPER messy...


Bigfoot71:
Everything is fixed, I'm merging!

@grunthon ping if you want to take a look

grunthon:
I'll go over this more closely later. That matrix ordering issue with the instanced transforms is a sneaky one.

Issue #96, opened 2025-09-29
Possible improvements to rendering pipeline


grunthon:
I've done some work on improving forward rendering (like precomputing normal data, ssao) and it got me thinking about how everything flows in both deferred and forward pipelines.

If seems like having both forward and deferred rendering output the same data to the render target textures could avoid a lot rendering method specific code. The main change would be to have forward rendering output the same separated color data (diffuse, specular, etc) as deferred. Then the scene shader could create the final output regardless of how it was rendered. This would also always give any post processing access to the same data. Potentially adding a separate render texture for ambient light could also let SSAO work the same for both without having to precompute normal data for forward rendering.

Of course there are potential downsides I won't get into at the moment, but it does have the advantage of unifying a lot of functionality. Looking for feedback on the general concept, or see if there are any major flaws that I'm missing.

Bigfoot71:
Yeah, it's a kind of "hybrid forward MRT"? The issue with that is transparency (surely what you imply), it would totally work for opaque rendering, but for transparency you’d still need a "pure" forward pass on top, otherwise it seems to me that everything behind would end up poorly/no shaded

Bigfoot71:
Well, even if it won’t work for transparency, if it’s okay (it seems so off the top of my head), it’s 1000 times better than a depth pre-pass to be able to support SSAO in forward

grunthon:
I'll go ahead and do some testing soon. I'll probably start with seeing about getting SSAO working for forward without having to do a prepass, and moving SSAO to post-processing.

And yeah, dealing with transparency is always going to be kind of messy with this kind of setup. I'll think more about all of this, and let me know if you have any other thoughts.

Bigfoot71:
> I'll go ahead and do some testing soon. I'll probably start with seeing about getting SSAO working for forward without having to do a prepass, and moving SSAO to post-processing.

Just wait a bit beforehand, _just in case_, I'm making major simplification changes to the shaders and r3d_drawcall (stupid name in hindsight), if you start before you might struggle later

I’ll open a PR and mention you if you want, I should be done by tomorrow evening at the latest, but hopefully tonight if I’ve got the mojo

grunthon:
No problem or rush. I won't have time to work on it until tomorrow at the earliest.

Bigfoot71:
Since opaque rendering in forward has been abandoned, I'm closing this issue

PR #99, opened 2025-09-30
SSAO in post-process

grunthon:


grunthon:
I worked on this a bit more but still have a couple things to figure out. Particularity with being able to apply SSAO to ambient and diffuse separately in both forward and deferred. I should have time tomorrow to finish it and it can be reviewed.

grunthon:
So I was looking over this again and here's where it is, particularly regarding SSAO.

I currently have the ambient render target holding only the ambient light value, unaffected by surface color. This could be used to attenuate and add SSAO in a separate pass at the start of post-processing and work fine. I had the thought that this could also be changed to essentially be a SSAOaffect specific texture, which stores how much the combination of ambient and diffuse should affect SSAO. This would also allow for per-material(and global) SSAO diffuse affect since it would be calculated during lighting when we have access to the materials.

The other option is to just go ahead and try the render target parity for forward and deferred discussed in #96. That would let the scene shader do the final light composition with SSAO for both rendering methods. This on it's own wouldn't really solve per-material SSAO, and lead to other changes needing to be made outside of just SSAO.

I'll probably go ahead and try the first idea sometime soon unless something better comes up. Let me know if there are any thoughts or ideas on this.

Bigfoot71:
I've gone through all of this carefully, and yeah, clearly there's no other choice than to store light affect per material in screen space.

No matter how you approach it, given how things are going, it will end up in some kind of forward + MRT setup anyway.

Right now, if we keep the forward shader as it is, even if we store ambient and diffuse in a buffer, there's no way to apply SSAO's contribution independently to ambient and diffuse if the target is the already shaded scene. So we'll inevitably end up with an opaque forward shader that is basically `geometry.vert/frag` plus lighting into a light buffer.

Even then, we'd need a separate buffer for ambient and direct lighting, because otherwise we'd end up mixing both in the same buffer, effectively doubling memory usage. And if we also need to store light affect per material in yet another buffer, it just keeps piling up.

Let me know if I misunderstood any part of what you're trying to do, but this seems way more complicated on closer inspection.

The other option would be a depth pre-pass for opaque forward to get depth and normals first, compute SSAO from that, and then apply it in forward. But that requires two vertex shader invocations, which is quite heavy if there are animations in the scene.

The problem with the forward + MRT approach is that, on mobile, it ends up having the same pitfalls as deferred anyway. I initially added opaque forward only for GL ES, which is pretty tricky with r3d, hence why I made Hyperion, which is natively GL/ES compatible.

Moreover, after a lot of research, it turns out that on mobile SSAO is very often just applied on top of the scene without considering ambient or anything else, because it becomes too technical for that hardware. That's actually why I think Godot simply didn't implement SSAO for their mobile pipeline.

So, after thinking this through carefully, it seems we'll have to live with some disappointment, unless I misunderstood something or you have some sort of black magic alternative in mind :')

grunthon:
Yeah, this has turned out more complicated than I first thought, and your assessment pretty much matches what I've come across. I did some more on it yesterday and the short version is that I'm now storing the ambient light contribution(affected by the albedo) in the new texture, which means that later we can essentially remove it after all lighting is done. This could also be combined with any amount of diffuse light we want SSAO to affect. It might seems a little odd at first, but isn't really that different from what would be happening with applying SSAO at any point.

If you want to see where it is right now, and maybe try it, I have a branch here https://github.com/grunthon/r3d/tree/testing
Try to ignore the testing stuff and some things happening is odd places(like the new render texture being clear in the background pass). It is generally working, with the exception of something wrong in non-IBL deferred, but I haven't really looked into that. Also some possible transparency issues.

I have noticed some other things that might need to be addressed (unrelated to what I've been doing). I'll try to keep track of any.

grunthon:
Quick note for this. I didn't intend to have this sitting here for so long, but as discussed it wound up more involved and I haven't had spare time to work on it. I should have time this weekend to get it to something that can be decided on whether to merge or not.

grunthon:
This is now complete to the point of accomplishing the original intent and could use a review. I can update the title and initial comment later but here are the relevant notes for now:

New shader to apply SSAO in post-processing and SSAO is now functional in both forward and deferred rendering.

SSAO is only being applied to ambient light, but this should make it pretty straightforward to have per-material SSAO included in the new buffer that is attenuating SSAO.

The `r3d_clear_additional_buffers` function is only a temporary solution to clear the new ambient texture since it is not included in the gbuffer. I tried different handling this in different places but it really needs to be cleared before any visible draw calls to work correctly. I think replacing `r3d_clear_gbuffer` with a more universal method to clear any needed buffers would probably be best, but I didn't really want to be making more large changes here.

Transparency with SSAO needs to be reviewed.

Naming of the added shader/buffer/etc will likely need to be changed to be more appropriate.

Bigfoot71:
I needed some time to think this through because I have quite a few doubts about this

Initially, I only considered SSAO for the forward path on GLES, but as I progressed with my other project, I started wondering whether it might be better to keep r3d strictly desktop-only, supporting mobile would require many changes since a lot of r3d relies on things that desktop drivers handle efficiently but not the mobile ones

That said, your approach is interesting, if I didn't have these doubts, I'd probably merge it right away

The only real issue (aside from transparency, which can likely be solved without too much trouble) is that in "ssao_apply" you clamp the result, and that's correct, because ambient must be applied first to avoid breaking transparency, but because of that, the formula `color -= (1.0 - ssao) * ambient` becomes incorrect

However, if we separate opaque and transparent shading (to avoid applying SSAO on transparent surfaces), it seems to me that we could simply not apply ambient during the opaque pass and instead apply it together with SSAO afterward. Then, in the transparent pass, ambient would be applied directly. That would probably solve both of those issues at once.

But even with that this approach isn't really ideal for mobile anyway, it might be an acceptable trade-off if someone decides to use SSAO on mobile they'll know exactly what performance cost they're paying, but honestly, we're still quite far from proper mobile support in r3d, and since SSAO for the forward path was originally meant for that specific purpose (and opaque forward in general), it makes me question whether it's still worth pushing in that direction.

So I'm sorry, but the real question I'm asking myself with this PR is do we really want to support mobile with r3d?

That said, your approach could absolutely be adapted in my other project, it already has GLES support, so we can be much more relaxed about it there.

grunthon:
I'm actually not familiar enough with GLES to really assess any potential issues without doing more research. It was not something I really kept in mind while working on this.

> The only real issue (aside from transparency, which can likely be solved without too much trouble) is that in "ssao_apply" you clamp the result, and that's correct, because ambient must be applied first to avoid breaking transparency, but because of that, the formula color -= (1.0 - ssao) * ambient becomes incorrect

Yeah, there was indeed a problem here. I was attempting to clamp the output color value to prevent it from possibly going below zero when was `uSSAOPower` is greater than 1, but this does not solve the problem of `(1.0 - ssao)` when `ssao` goes above 1 due to the power factor. I have changed this to clamp the `ssao` value itself instead of the output. There likely could be improvements to how this is handled, or other ways the post-process SSAO pass might work, but I won't bother with doing more when this might not end up being used.

One thing I did consider while working on this was just having a separate naive SSAO mode that would simply apply the effect to the scene after rendering without any regard to materials or the lighting source. This is pretty commonly done and avoids all the complications of needing information only available at earlier points in the pipeline, at the expense of "realism".

All this aside, if this doesn't seem like a good approach for the library or might conflict with any future development ideas, feel free to close this. It was really just an idea that came from #96 where the actual thinking was to bring forward and deferred render targets in line to make everything more unified.

Bigfoot71:
> Yeah, there was indeed a problem here. I was attempting to clamp the output color value to prevent it from possibly going below zero when was `uSSAOPower` is greater than 1, but this does not solve the problem of `(1.0 - ssao)` when `ssao` goes above 1 due to the power factor.

Ah that's not even what I had in mind, the issue I had noticed is that, on the shading side we calculate the scene color like this:
```
diffuse = albedo.rgb * (ambient + diffuse);
FragColor = vec4(diffuse + specular + emission, albedo.a);
```

So let's take these values (scalar for simplicity):
```
albedo = 0.2
ambient = 0.5

color = albedo * ambient => 0.1
```

Then we might get this:
```
ssao = 0.5
color -= (1.0 - ssao) * ambient
color => 0.1 - (1.0 - 0.5) * 0.5 => -0.15
```

That's why I thought you were clamping the result.

And that's why I said this:
> However, if we separate opaque and transparent shading (to avoid applying SSAO on transparent surfaces), it seems to me that we could simply not apply ambient during the opaque pass and instead apply it together with SSAO afterward. Then, in the transparent pass, ambient would be applied directly. That would probably solve both of those issues at once.

Bigfoot71:
Honestly, we can adjust this and merge, it does the job for SSAO with forward rendering, my main interest was to have SSAO on mobile with a full forward pipeline, but since I’m reconsidering that for r3d

If you see other advantages to doing opaque rendering in forward on desktop, let me know, that would be perfectly fine for me as well

grunthon:
The albedo is already applied to the ambient light value stored in the new texture so it shouldn't be possible to remove more than it previously contributed (outside of the case I mentioned, or possible issues with transparency). Along the lines of your example this is what should be happening (assuming only ambient light is being used to keep it simple)

```
ambient = 0.5
color = 0.5
```

```
ssao = 0.5
ssaoColor = (1.0 - ssao) * ambient = 0.25
FragColor = color - ssaoColor = 0.25
```

This is just me thinking about it off the top of my head, and I could be missing something in either what you're saying or how I have it working right now. I can actually look over it later.

Related to this is that the albedo was being pre-multiplied into ambient light in non-IBL deferred, which happens later in the scene shader when the light is combined.
https://github.com/Bigfoot71/r3d/blob/2e24e38f761c7de0b9104af75e0baa029e51ee64/shaders/screen/ambient.frag#L182

Bigfoot71:
Ah yes sorry, you're right, I must be too tired, you're doing it in forward.frag:
```glsl
FragAmbient = albedo.rgb * ambient;
```

Anyway, I tested it and the only visual issue I saw was on the foliage in the Sponza example, but that will be fixed if we render transparent objects after the SSAO

So for me, we can merge, and I can take care of the transparency issue, unless you want to handle it first

Whether I add GLES support or not (I might take some time to decide), at least we'll have this part done

And it doesn't conflict with anything, it's just that forward opaque (thus ssao here) is not really useful in practice if we're not targeting mobile imo, but it will be done if we need it

grunthon:
As long as you're sure about merging, feel free to handle the transparency stuff. I just wouldn't want you to merge if you're unsure about it or think you might want to do more testing first.

Getting a per-material SSAO affect added after this shouldn't be too difficult, which is something that would be an improvement even beyond just trying to have more parity between forward and diffuse rendering.

grunthon:
I was checking back on this to see if I missed any updates. Rereading my last comment I noticed it was worded in a way that might have sounded like there was more that needed to be done before merging, which is not what I was meaning to say. I was just trying to point out that there is more that can be done surrounding this if it does gets merged.

No problem if there just hasn't been time for this, but I wanted to make sure there wasn't a misunderstanding.

grunthon:
I'll probably go ahead and close this soon. It's been sitting here for too long and I've since thought that just breaking apart light data for both forward and deferred and combining lighting(including SSAO) for both in the scene shader would probably be better.

These are my raw notes I had for how this would work:

> Change "framebuffer.deferred" to a general "framebuffer.render" used by both forward and deferred.
> Break diffuse into separate ambient/diffuse textures.
> Add an additional material data texture to store material's SSAO strength values for ambient/diffuse (it could be possible to pack this data into another channel of the ORM texture and avoid an additional texture).
> 
> Render to "framebuffer.render"
> deferred
> forward opaque
> forward transparent(sorted)
> (both forward and deferred output ambient, diffuse, specular, emission, albedo, normals, depth, orm, SSAO/material data)
> 
> Calculate SSAO.
> 
> Scene shader then combines all lighting and applies SSAO based on material data texture and/or global settings.

Let me know if you were still considering merging this PR, otherwise I'll close it and sometime soon see how this other approach works out. As always, don't be shy about letting me know if you don't like any of this, don't think it's needed, or would just prefer to work it out yourself.

Bigfoot71:
Hey, I haven't been connected on github for a while and missed that the PR got closed, I'm really busy right now and I’m fighting to make progress where I can, so really sorry if you were expecting a quick reply.

I didn't merge right away because I needed some time to think things through. Working on my other project made me rethink r3d, I've got a list of fixes and improvements and I'm thinking it might be better to drop mobile support since it would make things really more complicated here. Mobile stuff will go into my [other project](https://github.com/Bigfoot71/Nexium) instead.

Without mobile, opaque forward rendering isn't really needed, so SSAO for forward doesn't make sense if we're only doing non opaque rendering.

If you have any reasons to keep opaque forward outside of mobile I'm all ears but for me it just seems like extra complexity with no practical use.

Anyway, for me the whole question about the usefulness of all this was "should we support mobile with r3d?"

PR #101, opened 2025-10-20
remove incorrect albedo in ambient light

grunthon:
The ambient light calculated here goes into the diffuse render target which is used in the scene shader where albedo is applied. Applying it in the shader means that the albedo was affecting ambient light twice.


Bigfoot71:
Ah yeah, I just tried it and yep that's why it was all dark, thanks!

Issue #105, opened 2025-10-28
Question about model animation


DuyhaBeitz:
Do I have to load the same model for each object if I want to use animations? In raylib I can reuse the same model, but if I do that with r3d I get all the models playing the same animation. Am I doing something wrong or is animation data tied to the specific model?

DicoSky:
In Raylib, the animation data specifically the bone transforms  are stored inside the Model itself.

That means if you animate several objects using the same Model pointer, they’ll all share the same bone data and end up moving in sync.

The fix is to share the meshes and materials (so you don’t waste memory or GPU uploads), but make a deep copy of the bones and bind pose for each instance.
This way, each model has its own animation state while still reusing the same geometry efficiently.

Create a new model instance from the original (the original acts as a shared asset) :

`
//
Model loadModelFromAsset(Model source)
{
    Model target = { 0 };

    // Copy basic transform and counts
    target.transform = source.transform;
    target.meshCount = source.meshCount;
    target.materialCount = source.materialCount;

    // Share (shallow copy) mesh and material data
    target.meshes = source.meshes;
    target.materials = source.materials;
    target.meshMaterial = source.meshMaterial;

    // Deep copy bones and bind pose (if any)
    target.boneCount = source.boneCount;
    if (source.bones && source.boneCount > 0)
    {
        target.bones = (BoneInfo *)RL_CALLOC(source.boneCount, sizeof(BoneInfo));
        memcpy(target.bones, source.bones, source.boneCount * sizeof(BoneInfo));

        target.bindPose = (Transform *)RL_CALLOC(source.boneCount, sizeof(Transform));
        memcpy(target.bindPose, source.bindPose, source.boneCount * sizeof(Transform));
    }
    else
    {
        target.bones = NULL;
        target.bindPose = NULL;
    }
    return target;
}
`

Maybe someone else has a different approach, but this is how I do it.

grunthon:
There are currently limitations to how R3D handles animation. Animation data for each each unique `R3D_Mesh` to be drawn is generated when `R3D_End` is called, and then stored inside the mesh. Since this only happens once per frame, every time the same `R3D_Mesh` is drawn in a individual frame they will always share the same animation data. Only separate, unique meshes will have independently cached animation data.

[...]

PR #106, opened 2025-10-30
per drawcall animation update

grunthon:
Removes the single per frame animation update in `R3D_End` and replaces it with individual updates for each drawcall. I had already been doing this for some time to correctly handle drawing the same mesh multiple times in a frame while using different animation data.  This was recently brought up in #105.


Bigfoot71:
Thanks for the correction! There are quite a few corrections to be made to the animations anyway, I'd like to fix that while also making the model loading cleaner, I hope to have the time soon...

PR #109, opened 2025-11-04
Color and energy control for ambient light

grunthon:
Breaks standard ambient light into a color and energy value to give more control, along with treating it more as an actual light source.

I did set the default ambient lighting to zero. I don't think having added ambient light by default makes any more sense than adding a default light of any type. I did review the examples to make sure this wasn't causing any problems there.

I can see users getting a black screen and not understanding that it's due to there being no default lighting as a reasonable argument for having some default ambient light, but if nothing else I think the previous value was too high.

### DIFF
--- a/examples/basic.c
+++ b/examples/basic.c
@@ -25,6 +25,8 @@ const char* Init(void)
 
     /* --- Setup the scene lighting --- */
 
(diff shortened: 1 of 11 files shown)


Bigfoot71:
Sure, if we take it in terms of "pure lighting", if nothing is specified then there shouldn't be any, it's consistent! Let's go!

Issue #113, opened 2025-11-12
[Alpha] Is it a transparency sort problem ?


TokyoSU:
I'm trying to make it work :') i did enable transparency and opaque sorting.
<img width="1915" height="1110" alt="Image" src="https://github.com/user-attachments/assets/045028f8-b215-4ee5-8d38-f3be98bab222" />
[meshes.zip](https://github.com/user-attachments/files/23506024/meshes.zip)

TokyoSU:
Maybe a depth mask problem.
Does it need to be the last loaded to work ?

TokyoSU:
Yes it's because the model was not the last to draw 👍 
Sorry :')
<img width="907" height="661" alt="Image" src="https://github.com/user-attachments/assets/4f479804-c742-4e64-897a-72902e06006f" />

Bigfoot71:
No harm done, any issue can be useful to others!

And yes, for more complex cases, such as models with transparent and opaque meshes, you also have the `R3D_FLAG_TRANSPARENT_SORTING` flag

But if you can manage without it, that's better ;)

TokyoSU:
It was enabled but seem like it was also because it was drawing near the beginning of the list :')

PR #114, opened 2025-11-12
Added BLEND_PREMULTIPLIED_ALPHA and DEPTH_MODE

TokyoSU:
- Added BLEND_PREMULTIPLIED_ALPHA when drawing multiple alpha model that need it.
- Added R3D_DepthMode to control depth per mesh, default enabled on test and mask.

Did add these since a model of mine require it, maybe that will be useful for some people too :)

- Example disabled (or before implementation even with alpha or additive enabled):
Capture d&#39;écran 2025-11-12 200011
Capture d&#39;écran 2025-11-12 200023
Capture d&#39;écran 2025-11-12 200031
- Enabled:
Capture d&#39;écran 2025-11-12 195948
Capture d&#39;écran 2025-11-12 195942

### DIFF
--- a/include/r3d.h
+++ b/include/r3d.h
@@ -113,10 +113,11 @@ typedef unsigned int R3D_Layer;
  * @note The blend mode is applied only if you are in forward rendering mode or auto-detect mode.
  */
 typedef enum R3D_BlendMode {
-    R3D_BLEND_OPAQUE,          ///< No blending, the source color fully replaces the destination color.
-    R3D_BLEND_ALPHA,           ///< Alpha blending: source color is blended with the destination based on alpha value.
-    R3D_BLEND_ADDITIVE,        ///< Additive blending: source color is added to the destination, making bright effects.
-    R3D_BLEND_MULTIPLY         ///< Multiply blending: source color is multiplied with the destination, darkening the image.
+    R3D_BLEND_OPAQUE,               ///< No blending, the source color fully replaces the destination color.
+    R3D_BLEND_ALPHA,                ///< Alpha blending: source color is blended with the destination based on alpha value.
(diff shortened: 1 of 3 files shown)


TokyoSU:
if you need a model to test (or for a depth example ?): https://github.com/Bigfoot71/r3d/issues/113#issue-3617519375

Bigfoot71:
Looks good to me, thanks for that, I'm just wondering if it wouldn't be more flexible to put the depth mode in R3D_Material

Just a question, perhaps you have some arguments for R3D_Mesh, but in my other [project](https://github.com/Bigfoot71/Nexium), I ended up implementing all sort of depth control in the materials.

TokyoSU:
Yes i wanted to make it for mesh, but if material is okay i can test it :)
EDIT: I tried using material instead of mesh, please no :(
I did try it with this model (per material is not working):
Capture d&#39;écran 2025-11-13 173805
Capture d&#39;écran 2025-11-13 173831
And if it's not per mesh, i will not be able to set depth (writeonly) there:
Capture d&#39;écran 2025-11-13 173918

TokyoSU:
It's very specific in my case, but that allow more customization over depth if it's per mesh (even instanced mesh).
Or maybe i did apply it wrongly in the code, it was where the blend and cull mode was applied.

TokyoSU:
I did also check to load premultiplied_alpha but blender not support it (also assimp too), depth too is not supported by assimp/blender.
All these option need to be defined manually.

Bigfoot71:
Sorry for the delay, and yes, it’s not really something standard, but no problem, I’ll merge it as is
The depth mode will stay in mesh as long as no one complains about it

PR #116, opened 2025-11-27
initial implemention of decal rendering

grunthon:
_**Oops. Not sure how the other reverted commit from something else wound up in this. I can make a clean PR if wanted, but it doesn't affect anything.**_

I'm putting this up for review and possible merging, but there are some things to note.

I've removed drawing normal data for the decals in this PR. I've done quite a bit on this already but it still needs to be better and I will add it when I'm more satisfied with it.

The shaders are not really cleaned up right now and include unneeded uniforms. This is mostly from the ongoing work with normals and will be dealt with when that is finished.

Sorting of draw order needs improvement for overlapping decals. Ideally the decals should be drawn on top of each other in the order of the draw calls but I've had issue with this so more work needs to be done on this.

Angle rejection. Oblique angles give overly stretched textures. Being able to define a cutoff or fade angle will help avoid this.

There needs to be a way to have objects not be affected by decals. For example a character moving in front of a wall with a decal on it should probably not have the decal in it. I already have ideas for how to handle this.

I started to create functions along the base/ex/pro concept used for other draw functions, but the transformation of decals is pretty central to using them correctly so I'm unsure about trying to simplify it. Perhaps a base function that takes position, rotation axis/angle, and scale in addition to a pro one that takes a transform matrix.

It's debatable if decal functions and shaders should be defined as screen rather than raster, they are somewhat both. Eh.

All this aside, this is completely functional for basic decal rendering. Any thought or feedback is welcome. If this can be merged, I will work on improvements soon.


shaders/raster/decal.frag
```diff
@@ -0,0 +1,98 @@
...
+uniform mat4 uMatInvView;
+uniform mat4 uMatNormal;
+uniform mat4 uMatVP;
+uniform mat4 uMatInvProj;
+uniform mat4 uMatProj;
+
+uniform vec2 uViewportSize;
+
+uniform float uAlphaCutoff;
+uniform float uNormalScale;
```
Bigfoot71: Perhaps add a comment saying "unused, will be used later"?

shaders/raster/decal.vert
```diff
@@ -0,0 +1,70 @@
...
+layout(location = 2) in vec3 aNormal;
+layout(location = 3) in vec4 aColor;
+layout(location = 4) in vec4 aTangent;
+
+layout(location = 10) in mat4 iMatModel;
+
+/* === Uniforms === */
+
+uniform mat4 uMatInvView;
+uniform mat4 uMatNormal;
```
Bigfoot71: Same as above, a comment?

src/details/r3d_drawcall.c
```diff
@@ -651,12 +729,12 @@ void r3d_drawcall_instanced(const r3d_drawcall_t* call, int locInstanceModel, in
     switch (call->geometryType) {
     case R3D_DRAWCALL_GEOMETRY_MODEL:
         r3d_drawcall_bind_geometry_mesh(call->geometry.model.mesh);
+        r3d_drawcall_apply_depth_mode(call->geometry.model.mesh->depthMode);
```
Bigfoot71, line 728: Wow, it's cool that you fixed that x)

Bigfoot71:
Hey, sorry I didn't get around to this sooner. I read all the changes and didn't see any obvious problems.

Aside from what you mentioned about the normals, maybe a small line to say it's unused but will be needed later, but I'm happy to merge it if that works for you!

And thank you for that! :D

grunthon:
I did some cleanup in the shader and marked the placeholder values. I was going to work on this some more soon which might result in API changes, but I don't think that really matters at this point.

include/r3d.h
```diff
@@ -396,6 +396,18 @@ typedef struct R3D_Model {
 
 } R3D_Model;
 
+/**
+ * @brief Represents a decal and its properties.
+ *
+ * This structure defines a decal that can be projected onto geometry that has already been rendered.
+ *
+ * @note Decals are drawn using deferred screen space rendering and do not interact with any
+ * foward rendered or non-opaque objects.
```
TokyoSU: forward* :p

src/r3d_core.c
```diff
@@ -695,6 +709,56 @@ void R3D_DrawModelInstancedPro(const R3D_Model* model,
...
+
+    drawCall.transform = transform;
+    drawCall.material = decal->material;
+    drawCall.shadowCastMode = R3D_SHADOW_CAST_DISABLED;
+    drawCall.geometry.model.mesh = &R3D.misc.meshDecalBounds;
+    drawCall.geometryType = R3D_DRAWCALL_GEOMETRY_MODEL;
+    drawCall.renderMode = R3D_DRAWCALL_RENDER_DEFERRED;
+    drawCall.depthMode = R3D_DEPTH_READ_ONLY;
+
+    r3d_array_t* arr = &R3D.container.aDrawDecals;
```
TokyoSU: could that be passed directly:  r3d_array_push_back(&R3D.container.aDrawDecals, &drawCall); ? (same for instanced)
}

include/r3d.h
grunthon: Foward rendering is my new highly advanced technique. It's faster since it has less letters. ;)

src/r3d_core.c
grunthon: I suspect it doesn't make any difference to (optimized) compiler output, but yeah there's no need for the extra pointer. I think I meant to get rid of it but just forgot to. Thanks.

Bigfoot71:
> I did some cleanup in the shader and marked the placeholder values. I was going to work on this some more soon which might result in API changes, but I don't think that really matters at this point.

Yeah, anyway as long as it’s not in release, nothing is guaranteed

So I’m merging it, because in any case it will be done, and I’ll rework the whole model/skinning part and their loading this weekend _**(FINALLY!)**_

At least this way you’ll avoid issues if there are any conflicts _(I don’t think there will be, but hey)_

Thanks for that!

Issue #98, opened 2025-09-30
Incorrect bind pose application on skinned meshes


Bigfoot71:
There's currently a skinning issue, as soon as an animation is applied, the model gets shrunk.

What's happening is that vertices are always transformed by the bind pose, even though that should only happen for static meshes. For meshes tied to a skeleton, vertices shouldn't be pre-transformed the animation should be applied directly. And if no animation is active, we should just keep the bind pose aside and apply it only in that case.

That's how I had implemented it in Hyperion: https://github.com/Bigfoot71/Hyperion/tree/master/source/Render/Core/PoolModel

Just leaving myself this note for later. Also, it would be nice to modularize all the model loading logic at some point, because it's a total mess right now. _(If anyone wants to tackle it before me, I won't complain :p)_

Bigfoot71:
Fix: https://github.com/Bigfoot71/r3d/pull/119

PR #119, opened 2025-12-13
Complete Refactor of `R3D_Model`

Bigfoot71:
This PR will introduces a major overhaul, focusing on 3D model management and the overall organization of the library. The main goal is to make the API clearer and more powerful, especially for animations and mesh manipulation.

## New and Updated Types

The refactor introduces or updates the following types:

- `R3D_PrimitiveType`
- `R3D_MeshData`
- `R3D_Mesh`
- `R3D_BoneInfo`
- `R3D_Skeleton`
- `R3D_AnimationTrack`
- `R3D_AnimationChannel`
- `R3D_Animation`
- `R3D_AnimationLib`
- `R3D_AnimationState`
- `R3D_AnimationPlayer`
- `R3D_Model`

### Key Changes

1. **Animations:**
   The new API will greatly simplify animation management and make it more complete. Support for animation blending is also planned.

2. **Meshes:**

   * `R3D_Mesh` now only stores GPU handles and rendering parameters.
   * RAM storage of mesh data is no longer mandatory. Use `R3D_MeshData` if you need to keep mesh data in memory.
   * New functions will allow attribute generation (UVs, normals, tangents) and mesh manipulation (transformations, merging) through `R3D_MeshData`.
   * Support for primitive types will also be added to meshes, their rendering will no longer be limited to triangles lists.

3. **Skeletons and Skinned Models:**

   * Introduction of `R3D_Skeleton` to represent mesh skeletons.
   * Major refactor of skinned model handling.
   * Fix for the scaling issue on skinned meshes: https://github.com/Bigfoot71/r3d/issues/98.

4. **Model Import:**

   * Complete refactor of assimp model importing.
   * Planned improvements to material loading speed.

5. **Library Organization:**

   * Better module separation.
   * Split of the `r3d.h` header into multiple module-specific headers for better readability.
   * Each type will have its own header with declarations and associated functions.
   * Initialization and rendering functions centralized in `r3d_core.h`.


Bigfoot71:
## API Changes

* `R3D_SetImportScale` has been removed. Scaling should now be applied directly in a 3D editor (e.g. Blender). This feature was unreliable and not consistently supported across all model formats.

* `R3D_UpdateMeshBoundingBox` and `R3D_UpdateModelBoundingBox` have been removed.
  Since `R3D_Mesh` no longer stores CPU-side data (GPU handles only), bounding boxes must now be computed using
  `R3D_CalculateMeshDataBoundingBox` on `R3D_MeshData`.

* `R3D_GenMeshXXX` functions no longer take the `upload` parameter.
  Mesh data is now generated in RAM using `R3D_GenMeshDataXXX`, and GPU upload is explicitly handled later via `R3D_LoadMesh`.

* `R3D_DepthMode` has been moved from `R3D_Mesh` to `R3D_Material`.

* `R3D_LoadModelAnimations` has been renamed to `R3D_LoadAnimationLib` (see new animation system below).

* Setting `R3D_Layers` to `0` now explicitly means *do not draw*. This behavior was previously unclear.

---

## New Animation API

The previous animation system was simple but restrictive: animations were precomputed at a fixed framerate, frame-based control was manual, poses were recalculated every draw call, and animation blending was not practical.

The new API introduces a clearer separation of concerns and fully runtime-evaluated animations.

### Key Concepts

* **`R3D_Skeleton`**
  Represents a model's skeleton. This improves conceptual clarity and allows for future features such as dynamic skeleton swapping.

* **`R3D_AnimationLib`**
  A container holding all animations of a model, replacing the old system that returned a raw animation array at load time.

* **`R3D_AnimationPlayer`**
  Responsible for evaluating animations. It is created from a `R3D_Skeleton` and an `R3D_AnimationLib`, and contains one `R3D_AnimationState` per animation.

### `R3D_AnimationState`

```c
typedef struct R3D_AnimationState {
    float currentTime;
    float weight;
    bool loop;
} R3D_AnimationState;
```

Each state defines:

* the current animation time,
* its blending weight,
* whether the animation loops.

When calling `R3D_UpdateAnimationPlayer(player, dt)`:

* animation timers are updated,
* the final pose is computed based on weights,
* animation blending is applied automatically.

⚠️ It is recommended that the sum of all animation weights equals `1.0f`.

### Benefits

* No fixed framerate required at load time.
* Runtime interpolation and animation blending.
* Poses are evaluated once per update and reused for rendering.
* Full control over animation update frequency.

### Advanced Control

For finer control, time advancement and pose calculation can be handled separately:

```c
void R3D_AdvanceAnimationPlayerTime(R3D_AnimationPlayer* player, float dt);
void R3D_CalculateAnimationPlayerPose(R3D_AnimationPlayer* player);
```

`R3D_AnimationState` timers can also be updated manually for custom animation synchronization strategies.

grunthon:
I plan to look over this closer soon, but it's looking good. Let me know if there is anything in particular you might want a second set of eyes on.

Bigfoot71:
I think it's looking good too, except the `R3D_Sprite` rendering is broken...

It shouldn't be too hard to fix, but I was actually considering removing it anyway.
See: https://github.com/Bigfoot71/r3d/discussions/120

I'd be interested in your thoughts on that.

---

Also, I added parallel material texture loading using cthreads.
The sponza scene now loads almost instantly on my machine, compared to ~2 seconds before.

It's a first iteration, the job management will be improved, but the strategy itself might not be ideal for everyone.
See: https://github.com/Bigfoot71/r3d/blob/model-overhaul/src/details/importer/r3d_importer_texture.c

_Note that the real bottleneck now is the ORM texture composition_

---

Regarding decals, I haven't touched them. There's a header `r3d/r3d_decal.h` that only contains the struct, and I added a note suggesting that we could use `r3d_primitive_bind_and_draw_cube` instead of creating an `R3D_Mesh` for rendering.

See: https://github.com/Bigfoot71/r3d/blob/model-overhaul/src/r3d_state.h#L344

But I also plan to revisit (later) and possibly remove `r3d_drawcall.h`, simplifying it as much as possible and putting everything necessary directly into the new `r3d_draw.c`, so I can handle that at the same time.

---

Sorry for the long message, I'll run a few more tests tomorrow and tweak some things before thinking about merging

grunthon:
Ah I guess the multi-threaded texture loading is the reason for the troubles with MSVC/windows you seem to have had.

I actually was using `r3d_primitive_bind_and_draw_cube` for the decals but had to change it at some point. I think something with depth checking was giving me problems (from #114?). I'd have to check my notes. I've done more on the decals but was going to give them another pass after this was merged. They are a bit of a weird fit for `R3D_Mesh` and `r3d_drawcall_t`. I already had some ideas for improvements that would require more substantial changes.

Bigfoot71:
> I think something with depth checking was giving me problems (from #114?)

Yes, I saw that you're basing it on depth mode, because it's applied directly in `r3d_drawcall()`, which isn't very flexible in hindsight, we should be able to apply it manually outside of it.

Also after rereading the primitives code the VAO setup looks a bit odd, I’ll cleanup that after this PR.

___

> Ah I guess the multi-threaded texture loading is the reason for the troubles with MSVC/windows you seem to have had.

Yeah, apparently C11 atomics are still experimental in MSVC. You have to enable `/experimental:c11atomics`

I also needed to get the available thread count via `GetSystemInfo`, but it’s only usable through `windows.h`, which causes naming conflicts with raylib.

Seems fixed now, I’ll test it for real before merging, but my windows 11 is on a painfully slow HDD, so I’m relying on github actions for build check x)

_maybe investing some time in a VM would pay off_

grunthon:
> I also needed to get the available thread count via GetSystemInfo, but it’s only usable through windows.h, which causes naming conflicts with raylib.

Using `getenv("NUMBER_OF_PROCESSORS")` to get the logical core count could be an option to avoiding using `windows.h`.

Bigfoot71:
Yes, that's something to keep in mind if any issues arise, but it should be fine!

I’ve tested all the examples on Windows and everything works, so I'm merging it!

Note: Only the sprites aren't fixed yet but I'll remove them right now

PR #122, opened 2025-12-17
[internal] Split of `r3d_state.h`

Bigfoot71:
This PR aims to break up the messy `r3d_state.h` _(and some stuff from `details`)_ into several independent modules.

The main idea is to use lazy initialization wherever possible, since many shaders, render targets, and some internal resources don't always need to be loaded depending on the use case.

There was already an attempt to load resources on demand, but the logic was scattered across multiple files, which made it "error prone".

I started with the shader management module. Now everything is centralized in a single module, from storage to creation, usage, and unloading.

I organized it so adding a new shader is straightforward. Add its struct, then its loading function, define it in the implementation, add its unload call, and that's it, you can start using it. All shaders will be loaded automatically on the first call to `r3d_shader_enable()`

This lazy init system also will allows removing environment getters/setters and integrating the `R3D_Environment` structure more cleanly.

Feel free to share your thoughts!


Bigfoot71:
Typo in commit, it's just one module for now

Bigfoot71:
These functions have been removed from the API:
- `R3DAPI Texture2D R3D_GetBufferColor(void);`
- `R3DAPI void R3D_DrawBufferSSAO(float x, float y, float w, float h);`
- `R3DAPI void R3D_DrawBufferBloom(float x, float y, float w, float h);`

They were removed due to the need for adaptation following the introduction of the new render target module, as they were not worth maintaining.

Note that the new module managing render targets can very easily allow any internal buffer to be redirected to the screen during the blit, so it is likely that all other internal buffer rendering functions will also be removed.

Bigfoot71:
Also, during the rewrite I realized that the bloom had become a big incoherent mess, now everything has been fixed and is much clearer, and the results really show!

Bigfoot71:
**API update (lighting):**

- Added a new internal lighting module (`modules/r3d_light.c`), replacing the previous implementation in `details/r3d_light.c`.

- Removed:
  - `R3DAPI bool R3D_HasShadowMap(R3D_Light id);`
  - `R3DAPI void R3D_SetSceneBounds(BoundingBox sceneBounds);`

- API changes:
  - `R3D_Light` is now an `int32_t` instead of `uint32_t`.
    Negative values are used to represent invalid lights.
  - `R3D_SetLightRange()` now also applies to directional lights and controls the visible shadow distance.
    This replaces `R3D_SetSceneBounds()`, with the shadow projection now automatically centered around the camera.
  - `R3D_DisableShadow()` no longer takes a `destroyMap` parameter.
    Shadow maps are kept internally and may be reused by other lights.

Bigfoot71:
@grunthon This commit revisited the decals a bit: https://github.com/Bigfoot71/r3d/pull/122/commits/8cbacfe095cf130640e2e9542b2c7dd5980679c9

The primitives are now used instead of the cube mesh that was created for
And I added instance colors as an attribute

I know the instanced draw function doesn't support them yet, but it's there for later
Since the vertex shaders might all be merged soon, this will help avoid problems

grunthon:
> @grunthon This commit revisited the decals a bit: https://github.com/Bigfoot71/r3d/commit/8cbacfe095cf130640e2e9542b2c7dd5980679c9

I'll look over it. I already have improvements to decals (normal mapping and controls for angle cutoff) but I was kind of waiting for you to finish the refactoring, plus I haven't had much spare time. And yeah, the cube mesh was a bit of a hack. Good that you switched it.

> Since the vertex shaders might all be merged soon, this will help avoid problems

I'm curious about idea of merging the vertex shaders.

Bigfoot71:
> I'm curious about idea of merging the vertex shaders.

The idea is that most scene vertex shaders are almost identical, so we could have a single one, optionally compiled multiple times using `#ifdef`

It should make modifications easier, I'll do it and see whether it causes any complications elsewhere
But it's not planned for right away

Bigfoot71:
Alright, I think this is good. I'll reread everything in a few days before merging.

Note that in the latest changes, all environment getter/setter functions have been removed and replaced by a globally definable structure system.

See: https://github.com/Bigfoot71/r3d/discussions/117

However, individual states can still be retrieved/modified without overriding everything using these macros:
```c
R3D_ENVIRONMENT_GET(member)
R3D_ENVIRONMENT_SET(member, value)
```

Also note that each effect (bloom, ssao, fog, etc.) has its own struct, so it's somewhat like lego.

Additionally, the 'background/ambient' part in the new `R3D_Environment` might seem in a weird state currently, but I plan to work on reflection probes soon, which is why it's structured this way.

Also, the frustum bounding box culling functions have been removed (they were somewhat unnecessary).

___

_The craziest thing is that I achieved everything I set out to do for 0.6, and more, it exhausted me x)_

Bigfoot71:
In the latest commits, these two flags were removed:
- `R3D_FLAG_LOW_PRECISION_BUFFERS`
- `R3D_FLAG_8_BIT_NORMALS`

Currently, transparency handling is a bit shaky, and the `pbr_car.c` example is broken.
I will fix this by adding other transparency handling modes, rather than just "blend/opaque"

Aside from that, everything is fine. I'll merge to move on to the next tasks.

grunthon:
> However, individual states can still be retrieved/modified without overriding everything using these macros:
>
> ```
> R3D_ENVIRONMENT_GET(member)
> R3D_ENVIRONMENT_SET(member, value)
> ```

Is having these kind of macros in the API really a good idea? They seem likely to just create problems. I'd be interested in your thoughts about it, but felt it should be brought up.

Bigfoot71:
> Is having these kind of macros in the API really a good idea? They seem likely to just create problems. I'd be interested in your thoughts about it, but felt it should be brought up.

Yeah, I asked myself the same question... I was thinking that if someone needed that kind of shortcut it would only take two lines to write, but it also made the examples simpler, and it's completely optional.

So honestly, I don't know, in the end I figured it couldn't hurt.

PR #131, opened 2026-01-01
Add sRGB texture loading

Bigfoot71:
This addresses the first point on the roadmap for version 0.7

The main goal is to support sRGB textures, which will therefore be enabled by default for 8-bit albedo maps. I'll also be looking into compressed formats.

A parameter to disable this will be available, similar to the default texture filtering option.

Along the way, the necessary changes will give us full control over how textures are loaded and configured during loading.

One of the first benefits is that anisotropic filtering is now handled correctly (for some reason, raylib struggles with it and it seems to work randomly).

However this also means we'll have to manage more things ourselves, especially regarding compressed formats, something that will need careful thought given the potential future support for WebGL 2.

Anyway, here's an example of the current Sponza scene versus what it should actually look like

## Current

screenshot000

## Now (what it should look like)

screenshot000

### DIFF
--- a/src/importer/r3d_importer_texture.c
+++ b/src/importer/r3d_importer_texture.c
@@ -28,6 +28,8 @@
 #include <tinycthread.h>
 #include <stdatomic.h>
 #include <stdlib.h>
+#include <rlgl.h>
+#include <glad.h>
 
 #include "../details/r3d_image.h"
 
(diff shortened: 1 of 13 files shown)


src/importer/r3d_importer_texture.c
```diff
@@ -118,25 +120,194 @@ static bool ring_pop(loader_context_t* ctx, int* jobIndex)
 }
 
 // ========================================
-// TEXTURE WRAP CONVERSION
+// TEXTURE LOADING FUNCTIONS
 // ========================================
 
-static int get_wrap_mode(enum aiTextureMapMode wrap)
+static bool has_gl_ext(const char* name)
```
Bigfoot71, line 126: If we ever need to check other extensions, the best solution would be to iterate once and cache the results, or better yet, to be able to access the RLGL cache which already performs the exact search for each platform...

PR #132, opened 2026-01-02
Instance buffer

Bigfoot71:
This PR adds instance buffers!

In this change, it should be noted that I made the decision to completely remove `R3D_ParticleSystem`, which would have required too many additional changes for something that can easily be custom made.

So the public headers `r3d_particles.h` and `r3d_curves.h` have been removed.
I will rewrite the example `particles.c` to give a quick idea of how to make a basic one.

Then these functions were removed from `r3d_draw.h`:
- `R3D_DrawMeshInstancedPro()`
- `R3D_DrawModelInstancedPro()`

Here are the new instanced rendering functions:
```c
R3DAPI void R3D_DrawMeshInstanced(const R3D_Mesh* mesh, const R3D_Material* material, const R3D_InstanceBuffer* instances, int count);

R3DAPI void R3D_DrawMeshInstancedEx(const R3D_Mesh* mesh, const R3D_Material* material,
                                    BoundingBox globalAabb, Matrix globalTransform,
                                    const R3D_InstanceBuffer* instances, int count);

R3DAPI void R3D_DrawModelInstanced(const R3D_Model* model, const R3D_InstanceBuffer* instances, int count);

R3DAPI void R3D_DrawModelInstancedEx(const R3D_Model* model,
                                     BoundingBox globalAabb, Matrix globalTransform,
                                     const R3D_InstanceBuffer* instances, int count);

R3DAPI void R3D_DrawDecalInstanced(const R3D_Decal* decal, const R3D_InstanceBuffer* instances, int count);
```

And yeah, it still has to be provided via pointers in the functions, it's for consistency, but I'll fix that, it's another thing inherited from the very first versions which are now inconsistent and just restrictive...

It should be noted that the `Ex` functions will also be revised in a future PR, as I have in mind to add a `BeginCluster()` / `EndCluster()` system (or something like that) which will naturally replace the `globalAabb`. The idea is to provide a more advanced interface for the frustum culling, but I will detail that in the next PR.

For the new type `R3D_InstanceBuffer`, it currently contains 4 buffers (position, rotation, scale, color). You can specify via flags which buffers you want to use, as well as a fixed capacity for now.

You then have an upload function `R3D_UploadInstances()` which takes the buffer, the type of data you will upload, an offset, a count (in number of elements), and the data.

You also have functions `R3D_MapInstances()` and `R3D_UnmapInstances()` to map and copy memory if you have less trivial data structures, or just for simplicity depending on the case.


Bigfoot71:
@grunthon Also, I had to revise the decal example. I quickly corrected it to fit the new system, but it can be improved if necessary. And I also replaced the cube rendering with raylib's `DrawCubeWires()`, as is now possible.

Also I had to modify the vertex shader, and I took the opportunity to directly provide `inverse(finalMatModel) * uView.invView` as varyings, rather than calculating it for each fragment.

grunthon:
No problem. I'd already rewritten a lot of the example anyway, and really there could probably be a better simple example. That pretty much just came out of me testing the decals.

A lot of the decal shaders will likely be rewritten as well depending on how they end up working, per the recent discussion https://github.com/Bigfoot71/r3d/discussions/108#discussioncomment-15387933.

PR #133, opened 2026-01-02
[BREAKING] API Refactor

Bigfoot71:
This PR introduces several breaking changes, mainly to the draw functions.

The goal was to fix some oddities inherited from the very first versions of R3D, mainly regarding pointers.
There has been no need for a long time to pass structures by pointer in draw functions, passing them by value works fine and is even safer.

From now on, simple structures are used as follows:
- By value if the parameter is required and the function does not modify the structure.
- By non-const pointer if the function modifies the structure or related data (even if it does not modify the members directly).
- By const pointer if the value is optional.

This should make the API more consistent, less error-prone, and easier to bind in other languages.

All `Draw` functions have been revised to ensure mesh rendering functions have signatures consistent with model rendering functions.

Also, `R3D_AnimationPlayer` is no longer returned via pointer when loaded.
The `R3D_AnimationPlayer player` member has been removed from `R3D_Model` to avoid awkward dependencies and complicated management.

Now to draw an animated model, use the `R3D_DrawAnimatedModelXXX` functions.
If you draw a skinned model with `R3D_DrawModelXXX` (without an animation player), or call `R3D_DrawAnimatedModelXXX` with an invalid animation player, the model will simply render in its bind pose.

The `globalAabb` parameters have been removed from instanced draw functions, as a "cluster" system is planned to allow more advanced culling strategies.

Also, a new function `R3D_SetDefaultMaterial()` has been added to set the default material, for example when loading 3D models.
Additionally, the `R3D_MATERIAL_BASE` definition has been added, representing the base default material, similar to `R3D_Environment`.

Here is the complete set of new draw functions:
```c
R3DAPI void R3D_DrawMesh(R3D_Mesh mesh, R3D_Material material, Vector3 position, float scale);
R3DAPI void R3D_DrawMeshEx(R3D_Mesh mesh, R3D_Material material, Vector3 position, Quaternion rotation, Vector3 scale);
R3DAPI void R3D_DrawMeshPro(R3D_Mesh mesh, R3D_Material material, Matrix transform);

R3DAPI void R3D_DrawMeshInstanced(R3D_Mesh mesh, R3D_Material material, R3D_InstanceBuffer instances, int count);
R3DAPI void R3D_DrawMeshInstancedEx(R3D_Mesh mesh, R3D_Material material, R3D_InstanceBuffer instances, int count, Matrix transform);

R3DAPI void R3D_DrawModel(R3D_Model model, Vector3 position, float scale);
R3DAPI void R3D_DrawModelEx(R3D_Model model, Vector3 position, Quaternion rotation, Vector3 scale);
R3DAPI void R3D_DrawModelPro(R3D_Model model, Matrix transform);

R3DAPI void R3D_DrawModelInstanced(R3D_Model model, R3D_InstanceBuffer instances, int count);
R3DAPI void R3D_DrawModelInstancedEx(R3D_Model model, R3D_InstanceBuffer instances, int count, Matrix transform);

R3DAPI void R3D_DrawAnimatedModel(R3D_Model model, R3D_AnimationPlayer player, Vector3 position, float scale);
R3DAPI void R3D_DrawAnimatedModelEx(R3D_Model model, R3D_AnimationPlayer player, Vector3 position, Quaternion rotation, Vector3 scale);
R3DAPI void R3D_DrawAnimatedModelPro(R3D_Model model, R3D_AnimationPlayer player, Matrix transform);

R3DAPI void R3D_DrawAnimatedModelInstanced(R3D_Model model, R3D_AnimationPlayer player, R3D_InstanceBuffer instances, int count);
R3DAPI void R3D_DrawAnimatedModelInstancedEx(R3D_Model model, R3D_AnimationPlayer player, R3D_InstanceBuffer instances, int count, Matrix transform);

R3DAPI void R3D_DrawDecal(R3D_Decal decal, Matrix transform);
R3DAPI void R3D_DrawDecalInstanced(R3D_Decal decal, R3D_InstanceBuffer instances, int count);
```

Apologies for this sudden breaking change, it's been long overdue.
Better late than never, these functions should now remain stable.

### DIFF
--- a/examples/animation.c
+++ b/examples/animation.c
@@ -54,9 +54,9 @@ int main(void)
 
     // Load animations
     R3D_AnimationLib dancerAnims = R3D_LoadAnimationLib(RESOURCES_PATH "dancer.glb");
-    dancer.player = R3D_LoadAnimationPlayer(&dancer.skeleton, &dancerAnims);
-    dancer.player->states[0].weight = 1.0f;
-    dancer.player->states[0].loop = true;
+    R3D_AnimationPlayer dancerPlayer = R3D_LoadAnimationPlayer(dancer.skeleton, dancerAnims);
+    dancerPlayer.states[0].weight = 1.0f;
+    dancerPlayer.states[0].loop = true;
 
     // Setup lights with shadows
     R3D_Light lights[2];
@@ -89,7 +89,7 @@ int main(void)
         float delta = GetFrameTime();
 
         UpdateCamera(&camera, CAMERA_FREE);
-        R3D_UpdateAnimationPlayer(dancer.player, delta);
+        R3D_UpdateAnimationPlayer(&dancerPlayer, delta);
 
         // Animate lights
         R3D_SetLightColor(lights[0], ColorFromHSV(90.0f * GetTime() + 90.0f, 1.0f, 1.0f));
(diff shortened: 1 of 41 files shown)


Bigfoot71:
Note that I set a limit of "no more than 256 bytes" passed as a function parameter, just to make sure it wasn't ridiculous.

The two functions that clearly exceed this limit are:
- `R3D_DrawMeshInstancedEx()`: 304 bytes
- `R3D_DrawAnimatedModelInstancedEx()`: 292 bytes

But I consider them acceptable in the end because these aren't functions meant to be called sequentially in large numbers.

And it should also be noted that compilers are very good at optimizing these cases in the majority of cases.

Issue #89, opened 2025-09-08
[REQUEST] Explicit shadow passes


DicoSky:
Hi everyone ,

I noticed that when we add models to an octree system, objects behind the camera get occluded. The problem is that there isn’t a separate draw list for each light, so we can’t use an octree per light. On large map, it would be better to render only what each light can see, instead of looping over thousands of entities to check which ones are inside the camera frustum.

I suggest adding this  (R3D_BeginLightDraw , R3D_EndLightDraw ):

Example : 

```
for(uint32_t i = 0; i < lightList.size(); i++)
{
    R3D_BeginLightDraw(i);

        std::vector<ENTITY> listM = queryOctree(.....)

        // Draw all objects that should cast shadows for this light

        for(uint32_t j = 0; j < listM.size(); j++)
        {
            R3D_DrawModel(&listM[j].model, listM[j].pos, listM[j].scale);           
        }

    R3D_EndLightDraw();    
}
```

Bigfoot71:
> I suggest adding this (R3D_BeginLightDraw , R3D_EndLightDraw ):

This is a real problem that I’m aware of, and it’s actually "funny" because in my many early attempts with r3d, that’s how I handled shadow maps. But it made "simple" scenes long to write (without any abstraction or anything), and I really wanted it to be simple and quick to prototype for anyone.

But yes, some effort should definitely go into a hybrid system, one that can be automatic or manual

grunthon:
It's possible I'm not completely understanding the issue, but this does seem at least partly a scene management issue which might be outside of a rendering library?

Camera frustum culling is good for optimizing out unneeded draw calls, but not really great for handling what portion of a large scene/level should be sent for rendering each frame. It sounds like an octree is already being used, which can handle choosing what models are in an active area(including only possibly casting shadows), then rely on frustum culling after that. I could be wrong, but I believe frustum culling is currently not applied to objects drawn for shadowmaps?

Not to say there couldn't be improvements to what is drawn per pass for lights/shadows/visible geometry. More directly control could certainly be helpful.

Bigfoot71:
Unless I’m over-interpreting, I’m aware of an issue with manual frustum culling where we can only blindly remove objects outside the frustum, even if they could still cast shadows onto the scene, and there’s currently no efficient way to handle this.

So when I saw “octree, culling, shadow” that’s what came to mind

[...]

PR #134, opened 2026-01-02
Implementation of clusters

Bigfoot71:
This PR adds support for clustered draw passes using a bounding box.

Two new functions are introduced:
```c
R3DAPI void R3D_BeginCluster(BoundingBox aabb);
R3DAPI void R3D_EndCluster(void);
```

They start and end a clustered draw pass. All draw calls submitted inside the pass depend on the cluster's visibility. If the cluster is outside the scene or shadow frustum, every draw call in it is skipped. This applies to both scene rendering and shadows.

This lets users organize their scenes to minimize individual frustum tests, while keeping full control over how clusters are defined.

It also solves the custom shadow culling problem mentioned here: https://github.com/Bigfoot71/r3d/issues/89
And that closes the discussion opened here: https://github.com/Bigfoot71/r3d/discussions/118

The alternative would have been to expose explicit shadow passes, but that would quickly become too complex, and it still wouldn't solve omni lights, which must render six times.

Clusters are a much cleaner solution.

The `R3D_FLAG_NO_FRUSTUM_CULLING` flag has been removed, since everything needed for proper culling is now available.

For the same reason, the `globalAabb` parameter was removed from instanced rendering. Clusters should now be used instead. If instancing is done outside a cluster, no frustum test is performed.

The only tricky case left is skinned models. For now, their frustum culling relies only on clusters, since their bounding boxes match only the bind pose. I left TODO notes about that.

Currently, if a model has a skinning texture, it is automatically treated as visible. Even though the bind-pose frustum test would technically work, I chose not to complicate things until we have a proper, general solution.

Finally, for now only one cluster pass can be active at a time, they cannot be nested.
This could be a nice improvement for the future, enabling even more powerful hierarchical culling.

### DIFF
--- a/examples/animation.c
+++ b/examples/animation.c
@@ -12,7 +12,7 @@ int main(void)
     SetTargetFPS(60);
 
     // Initialize R3D with FXAA and no frustum culling
-    R3D_Init(GetScreenWidth(), GetScreenHeight(), R3D_FLAG_FXAA | R3D_FLAG_NO_FRUSTUM_CULLING);
+    R3D_Init(GetScreenWidth(), GetScreenHeight(), R3D_FLAG_FXAA);
 
     // Enable post-processing effects
     R3D_ENVIRONMENT_SET(ssao.enabled, true);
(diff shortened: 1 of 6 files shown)


grunthon:
Interested to see how this works in practice.

Since you're actively working on this, instead of a PR I'll just note here that `r3d_light_shadow_should_be_upadted` has a typo and should be `r3d_light_shadow_should_be_updated`

Bigfoot71:
> Interested to see how this works in practice.

You can manage your "clusters" and associated lists yourself, and then perform the iterations, like this:
```
foreach cluster {
  BeginCluster(cluster.aabb)
    foreach cluster.object {
      Draw(cluster.object)
    }
  EndCluster();
}
```

Then you can slice your scene as you wish

> Since you're actively working on this, instead of a PR I'll just note here that `r3d_light_shadow_should_be_upadted` has a typo and should be `r3d_light_shadow_should_be_updated`

Aaaah, I might never seen it, thank you x')

PR #135, opened 2026-01-05
Reflection probes

Bigfoot71:
**Note:** Big PR! Hope my comment is structured enough!

## TL;DR

- `R3D_Cubemap` replaces old skybox system
- `R3D_AmbientMap` -> global IBL, flexible diffuse/specular generation
- `R3D_Probe` -> local lighting, flexible indirect/reflection selection, 2 update modes, shadows/interior options
- Cubemap arrays -> solves sampler limits but OpenGL 4.0+ compatible hardware
- Max active probes per frame: 4 (configurable)

_See at the end a quick demo._

## 1. Skybox and Cubemap

- `R3D_Skybox` has been completely removed
- New type: `R3D_Cubemap`
  - Can be used directly as a skybox
  - Contains only the cubemap texture and face size

## 2. Global IBL and `R3D_AmbientMap`

- `R3D_AmbientMap` is generated from a `R3D_Cubemap`
- Completely decouples skybox from global lighting
  - You can have a skybox without IBL or IBL without a skybox
- Choose to generate diffuse only, specular only, or both
- Updates:
  - `R3D_GenAmbientMap` -> creates the map from a cubemap
  - `R3D_UpdateAmbientMap` -> refreshes it from a new cubemap
  - `R3D_UnloadAmbientMap` -> frees textures when no longer needed

## 3. Local Lighting and `R3D_Probe`

- Like `R3D_AmbientMap`, you can choose
  - Indirect lighting only
  - Reflections only
  - Both

- Handles:
  - `R3D_Probe` is just an alias for `int32_t`
  - Negative value -> invalid handle

- Update modes:
  - `R3D_PROBE_UPDATE_ONCE`: updates only when state changes, and once after creation
  - `R3D_PROBE_UPDATE_ALWAYS`: continuous update every frame

- Additional modes:
  - **Shadows:** include shadows in the probe capture (disabled by default)
  - **Interior:** disables global IBL and skybox contribution for indoor probes

- **Limitations:**
  - No compute shaders (OpenGL 3.3 + Mac with raylib), so dynamic probes are less efficient than in big engines, but acceptable for now.
  - Probes are rendered forward -> 8 lights max by default, no decals, etc.
  - Max probes influencing the scene at once is **4** _(but can be increased via `R3D_SHADER_NUM_PROBES` in `src/modules/r3d_shader.h`)_.

## 4. Technical Details

Cubemap arrays are used for all IBL maps (ambient maps and probes).

This avoids exploding the number of samplers in shaders when multiple maps are used. The same approach will be applied to shadow maps soon _(I'll review the roadmap)_, solving the "16 samplers max" problem that affected some users.

But the most perceptive among you may have noticed it, this relies on `GL_ARB_texture_cube_map_array` (introduced with OpenGL 4.0), so with this, R3D is no longer 100% OpenGL 3.3, but choices have to be made. It remains compatible with all Macs from 10-15 years ago, hopefully no one with an Intel HD 2000 will want to run R3D x)

## 5. Other Changes

The "cache" module has been moved to `src/r3d_core_state.h`, which is a bit more consistent, and many smaller tweaks across thousands of line changes, most other changes are "minor" internal adjustments.

## 6. DEMO!

[Enregistrement d'écran_20260105_023216.webm](https://github.com/user-attachments/assets/eb3ee137-67ac-4a1e-86e5-63ed7392a11e)

screenshot000


Bigfoot71:
Oh, and I almost forgot, the IBL has also been completely revised and corrected, although there are still a few minor details I don't visually like in the specular, it's much better than before!

I'm not merging it right now, I'll do a review later.

grunthon:
This is really good stuff but there is something I am curious about.

I have to say I don't quite follow IBL being tied to ambient light. Ambient light is typically meant to represent a flat isotropic light contribution. IBL is generally done outside of ambient lighting specifically because it does generate light values with a direction and therefore the possibility to create specular highlights. The same concept would seem to follow for reflections/reflection probes as well. Also functioning the way is currently does removes any use of traditional ambient lighting when using IBL.

Maybe I'm missing some larger concept for all this but I would be interested to hear the thinking behind it.

Bigfoot71:
For the diffuse component, IBL is used to approximate ambient lighting from the environment, which in PBR is indirect diffuse, and it can't be purely isotropic unless the environment is perfectly uniform, and even then the irradiance varies with surface orientation. The simple ambient color is just a non-physical fallback.

Regarding reflections, there should indeed be a dedicated section in `R3D_Environment`. Personally I find the probes intuitive, but yes, calling the global IBL "ambient map" is confusing.

It was originally going to be "environment map", but it ended up under `environment.ambient`, which is also inconsistent. Renaming `environment.ambient` would have been a solution.

We could also decouple the two into something like a `R3D_ReflectionMap`, but I tried to keep the project intuitive so new users can get things working without digging too much.

I'm open to suggestions if you have ideas for clearer naming, and I'll keep thinking about it too.

grunthon:
> For the diffuse component, IBL is used to approximate ambient lighting from the environment, which in PBR is indirect diffuse, and it can't be purely isotropic unless the environment is perfectly uniform, and even then the irradiance varies with surface orientation.

In this case though IBL is functioning as direct, if diffuse, light from the image. To be considered indirect light it would by definition need to not originate from the initial source. I do get that it can still be used to simulate ambient light and some renderers (I think at least Godot and Unity) allow for an image to be used for generating ambient light, but even then IBL isn't specifically tied to ambient light.

> It was originally going to be "environment map", but it ended up under environment.ambient, which is also inconsistent. Renaming environment.ambient would have been a solution.

To me something directly descriptive of what it is, along the lines of `R3D_ImageBasedLightingMap`, would be appropriate. I don't really see any reason for it be narrowed to anything that precludes other potential uses.

> I'm open to suggestions if you have ideas for clearer naming, and I'll keep thinking about it too.

I should probably take another look over how all this fits together in the recent changes just to be sure I'm not misinterpreting anything and see if anything comes to me.

Bigfoot71:
> In this case though IBL is functioning as direct, if diffuse, light from the image. To be considered indirect light it would by definition need to not originate from the initial source.

Well, diffuse IBL accounts for all the bounced light, LearnOpenGL explains it well [here](https://learnopengl.com/PBR/IBL/Diffuse-irradiance).

So even if you have, for example, the sun in the map, the goal is not to calculate its direct light, you will still need a directional light in this case, the IBL will be used for lighting coming from the sky, bounces on buildings, or whatever, which are indirect.

The way it's done remains an approximation, but that's true for everything in the end.

> I do get that it can still be used to simulate ambient light and some renderers (I think at least Godot and Unity) allow for an image to be used for generating ambient light, but even then IBL isn't specifically tied to ambient light.

It's used for ambient or probes, globally. Using environment-based diffuse indirect light (whatever way you get it) as a replacement for the old ambient color hack is standard in modern engines.

UE4 called it [ambient cubemap](https://dev.epicgames.com/documentation/en-us/unreal-engine/ambient-cubemaps?application_version=4.27), and that ambient cubemap also handled reflections. There are also "sky lights", it's the same thing but more advanced.

Godot also has this as an [ambient light source](https://docs.godotengine.org/en/4.4/classes/class_environment.html#enum-environment-ambientsource), though Godot separates environment diffuse indirect and environment reflections into two categories, which I think makes more sense.

If we want a reason why engines do it this way, I see two reasons, first is probably historical, to avoid confusing users, second, it avoids having two options that "disable" each other, and that's what I would like too. It's more intuitive for an artist/user, since the effect on the scene is basically the same for a normal user, ignoring technical details.

For the probes, they are local, so they're separate from the global environment. The environment IBL is global, while the IBL from the probes are local and blended with the rest.

> To me something directly descriptive of what it is, along the lines of `R3D_ImageBasedLightingMap`, would be appropriate. I don't really see any reason for it be narrowed to anything that precludes other potential uses.

My take is that it's better to name things based on their use case or what the feature actually does. Exposing what it "really is" in the name only makes sense if it's actually helpful to the user.

And here, the term IBL applies equally to "ambient lighting" and "local probes", so naming an IBL type for only one of them seems a bit odd to me, and not necessarily helpful in the end.

> I should probably take another look over how all this fits together in the recent changes just to be sure I'm not misinterpreting anything and see if anything comes to me.

Yeah, no problem. Anyway I agree it can be improved, and if you or anyone finds it confusing, no matter why, it clearly means we should do better. I won't release until that's sorted.

Anyway, I'm finishing the roadmap and then I'll let it sit for a bit, it's already eating too much of my free time x')

PR #136, opened 2026-01-05
[internal] Review of uniform names

Bigfoot71:
I've mainly revised the sampler names, they now have the suffix 'Map' when they come from materials, and 'Tex' for everything else (consider internal textures).

I've also revised the names of a few other uniforms, such as `uMatVP` which is now `uMatViewProj`, just to make things a bit clearer.

### DIFF
--- a/shaders/deferred/ambient.frag
+++ b/shaders/deferred/ambient.frag
@@ -23,17 +23,17 @@ noperspective in vec2 vTexCoord;
 
 /* === Uniforms === */
 
-uniform sampler2D uTexAlbedo;
-uniform sampler2D uTexNormal;
-uniform sampler2D uTexDepth;
-uniform sampler2D uTexSSAO;
-uniform sampler2D uTexSSIL;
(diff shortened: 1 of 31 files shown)


Bigfoot71:
The only samplers I haven't renamed are the shadow ones, as I'll be revisiting those soon.

grunthon:
In [src/modules/r3d_shader.h](https://github.com/Bigfoot71/r3d/blob/master/src/modules/r3d_shader.h) there are a few place where `alignas` is being used on a whole struct like
`alignas(16) struct r3d_shader_block_env_probe`

Unless there's some weirdness I'm not aware of, I don't think you can use `alignas` on a type like `struct`.

_edit: I actually meant to put this comment on the previous PR, but I can at least say that MSVC does give a warning about this._

Bigfoot71:
Ah, that was just to be explicit in the idea, but anyway the alignment will be 16 in this case, so it can be removed

PR #139, opened 2026-01-06
enable support for hdr images in Raylib submodule

grunthon:
This could be a build option if that seems needed, but having IBL without HDR support in the included Raylib seems counterintuitive.

_I do still think that it would be a big improvement if any of skyboxes used in the examples could be HDR but I know there was concern about increasing the size of the library. I'd estimate it would add 2-3MB per image upgraded to an equivalent HDR image._


Bigfoot71:
Funny, I was thinking about it earler! I was also planning to review all the examples with better assets.

Thanks!

Bigfoot71:
As for the size, I’ve thought about it, it's probably best to recommend cloning with depth 1. And clearly, the biggest stuff isn't the skies, but the models. The repo's history is already screwed, so well, might as well go all in x)

grunthon:
> The repo's history is already screwed, so well, might as well go all in x)

I'm wary of eating someone's cooking if their kitchen seems too clean ;)

PR #141, opened 2026-01-07
[internal] Light module update

Bigfoot71:
The lighting module has been completely reworked.

Shadows are now handled using texture arrays, and light data is sent to shaders through UBOs.

One drawback of the new shadow system is that resolution is no longer configurable per-light. In practice, this setting shouldn't be exposed anyway, at best it could be managed through an atlas, which may be revisited later.

The major benefit is that this finally removes the sampler limit issues that occurred on some devices.

By default, shadow map sizes are 2048 for spot and omni lights, and 4096 for directional lights (these values are configurable through defines).

Using UBOs significantly reduces code complexity, as well as the number of checks and API calls.

Several issues in the lighting module were also fixed. On my machine, I observed about a ~20 FPS improvement in the `lights.c` example, and the startup slowdowns have completely disappeared.

---

This PR finally concludes the roadmap, however there are still a few minor internal adjustments to be made, which I will try to complete as quickly as possible, then I will allow a little more time before the release of 0.7

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -18,7 +18,7 @@
 * [x] **Skybox Generation Support**
   Add support for procedural skybox generation.
 
-* [ ] **Improve Shadow Map Management**
+* [x] **Improve Shadow Map Management**
   Instead of using one shadow map per light, implement an internal batching system with a 2D texture array and `Sample2DArray` in shaders.
   *Note: This would remove per-light independent shadow map resolutions.*
 
--- a/examples/animation.c
+++ b/examples/animation.c
@@ -62,13 +62,13 @@ int main(void)
     R3D_Light lights[2];
     lights[0] = R3D_CreateLight(R3D_LIGHT_OMNI);
     R3D_SetLightPosition(lights[0], (Vector3){-10.0f, 25.0f, 0.0f});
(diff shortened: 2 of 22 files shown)


Bigfoot71:
While doing this review, I noted that a little cleanup of what is sent to and consumed by the shader could be done later

grunthon:
I'm now getting heap corruption errors on exit which seem to originate from https://github.com/Bigfoot71/r3d/blob/cbd003f1f629c55ad85d41d0aeac1136d86abe5e/src/modules/r3d_light.c#L63

I haven't looked further into it, but maybe something with shadows isn't being freed in the correct order?

stack
stack

Bigfoot71:
All good, that must have been it, I had made a really silly last minute tweak in `r3d_shadow_pool_init()`...
I'll blame it on my lack of sleep x)

Fix here: https://github.com/Bigfoot71/r3d/commit/dc2b73e4819a794d5b29922e4bf99ae97e73c73b

It's surprising that even Valgrind doesn't give me inconsistent results, even for allocations...
Anyway, I'll also review the allocation logic then, but it should be fixed as it is.

grunthon:
Quick testing on my end isn't catching any writes outside of allocated memory now, so it does seem fixed.

PR #143, opened 2026-01-08
store tangents in normal render target

grunthon:
I felt it was worth making this a separate PR ahead of it's coming use in adding normal mapping to decals. I checked through the other shaders to see if it could cause any issues and everything looks ok. Let me know any feedback.

I could see there now possibly being a more accurate name for `R3D_TARGET_NORMAL` or at least some indication of how it is arranged to store both the normals and tangents.

Note this does make `R3D_DrawBufferNormal` behave in basically a broken way. This can be addressed if this way of storing the normals/tangents is alright to merge.

### DIFF
--- a/shaders/scene/decal.frag
+++ b/shaders/scene/decal.frag
@@ -43,7 +43,7 @@ uniform vec2 uTexCoordScale;
 
 layout(location = 0) out vec4 FragAlbedo;
 layout(location = 1) out vec4 FragEmission;
-layout(location = 2) out vec2 FragNormal; // Unused - normal output placeholder
+layout(location = 2) out vec4 FragNormal; // Unused - normal output placeholder
 layout(location = 3) out vec3 FragORM;
 
 /* === Main function === */
--- a/shaders/scene/geometry.frag
+++ b/shaders/scene/geometry.frag
@@ -36,7 +36,7 @@ uniform float uMetalness;
 
 layout(location = 0) out vec3 FragAlbedo;
 layout(location = 1) out vec3 FragEmission;
-layout(location = 2) out vec2 FragNormal;
+layout(location = 2) out vec4 FragNormal;
 layout(location = 3) out vec3 FragORM;
 
 /* === Main function === */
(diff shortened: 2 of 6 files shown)


Bigfoot71:
Yes, this should be fine. Everywhere we currently retrieve normals goes through `V_GetViewNormal()` or
`V_GetWorldNormal()`, we can add an equivalent helper for tangents as well.

For naming, yeah, `R3D_TARGET_NORMAL_TANGENT` or `R3D_TARGET_NORM_TAN` if
we want to be more concise would be more appropriate.

Regarding `R3D_DrawBufferNormal()`, I was planning to remove this kind of function
anyway and add a mode that lets us directly choose which output to display on
screen. So this isn't really an issue, and I can address it afterwards.

Let me know if you'd prefer to make the adjustments yourselves, or if I should
merge directly.

grunthon:
It is now `R3D_TARGET_NORM_TAN`. I considered changing the shader sampler names to reflect this but I'm not sure if that's really appropriate when it's only being used for the normals anyway.

Shader helper functions for tangents can be added but I'd think their main use (at least currently) is for texture mapping which isn't going to need transformation into other spaces. I guess they could be useful for SSR or other things in the future. The functions that only take a vector would be exactly the same, so just reusing them with a more generalized name is an option.

If you already have an idea for a new way to draw the buffers, I'll leave that to you unless you have specifics on it, in which case I have no problem doing it.

Bigfoot71:
> but I'm not sure if that's really appropriate when it's only being used for the normals anyway.

It would at least make it explicit, if someone comes across it, they'll know.
There's already `uTexORM`, which is sometimes used only for roughness and metalness, so it still makes sense.

> Shader helper functions for tangents can be added but I'd think their main use (at least currently) is for texture mapping

Yes, I just said that offhand, if we need it elsewhere later, sure no need to do it now.

> If you already have an idea for a new way to draw the buffers, I'll leave that to you unless you have specifics on it, in which case I have no problem doing it.

No need, I'm currently revising the blit modes on my side (with better upscale modes btw). I'll handle it along the way.

___

Just maybe the enum name for the sampler should be reviewed here as well:
https://github.com/Bigfoot71/r3d/blob/018ae8e12440c3c8e68bb7aa9b4b5f4e3d1e0600/src/modules/r3d_shader.h#L157

But it's fine as is. You tell me, and I'll hit the button!

grunthon:
The PR was mainly to see if the the idea was alright and get it in if so. Feel free to merge, unless you'd prefer to have dealing with sampler naming included in this.

PR #144, opened 2026-01-09
Blit review

Bigfoot71:
Last "big" PR from me for this version!

The idea behind this PR is that the previous upscale/downscale methods were... honestly terrible, and I needed an acceptable upscale mode for a project.

So I went through everything and separated the upscale and downscale logic:

- **Upscale**
  - Nearest
  - Linear
  - Bicubic (Catmull-Rom)
  - Lanczos-2

- **Downscale**
  - Nearest
  - Linear
  - Box

And when the internal and target resolutions match, it just does a simple nearest neighbor blit.

For upscale, [here's a comparison of each mode rendered at 1280x720 at 75%](https://imgsli.com/NDQwNjM1/0/1)

Honestly, 75% is already quite aggressive, I think the quality sweet spot is around 80% with FXAA and Lanczos.
For Sponza at 1280x720 with SSAO, I get 630 FPS on my RTX 3050, same setup at 80% with FXAA and Lanczos, I go up to 810 FPS.
That's a solid gain, and the visual differences are minimal, I really like it.

As for downscale, you might ask "why" but it can be useful in certain cases to serve as an SSAA (when it's acceptable, or simply for screenshots). That's why I even added a "box-blur" mode. It just reuses the dual-filter ARM downsampling shader from SSIL, but it gets the job done and is very effective. [Here's a small example of downscaling](https://imgsli.com/NDQwNjM3/0/1).

Well, the two examples shown in the images would look better in video, but hey, I'm too lazy x)

Finally, I cleaned up a few things:
- The render texture that R3D blits to, aspect ratio handling, etc, have been moved out of the target module, it's now managed by "core".
- The FXAA flag was removed and replaced with an enum, which will make it easier to expand later. I'd like to implement SMAA eventually, but honestly, not right now x)

That's it, I'm going to take a little break now.


Bigfoot71:
And I removed all the `R3D_DrawXXXBuffer()` functions, and added an enum `R3D_OutputMode` to indicate which buffer you want to see on the screen, via `R3D_SetOutputMode(mode)`

PR #147, opened 2026-01-12
Normal mapping and angle rejection for decals

grunthon:
Added functionality to decals
- Normal mapping
- Angle rejection based off deviation from surface normal
- Fading of edge along angle threshold

To allow for correct angle testing against the surface normals this also adds the new render target `R3D_TARGET_GEOM_NORM_TAN` which stores the base geometry's normals and tangents before any alteration from normal mapping (or other potential sources).

Adjustments have also been made to `r3d_draw_call_t` to properly handle decals and their accompanying data (thanks to @Bigfoot71).

<details>

<summary>Original draft comment for reference</summary>

I'm putting this up as a draft so it can be reviewed or checked out.

This first thing to note is that despite trying to have the storing of normals and tangents already addressed in #143, this changes the g-buffer normal back to it's previous state and there is now a new render target `R3D_TARGET_GEOM_NORM_TAN` that holds the raw geometry's normal and tangent data. The reason for this (and something I really should have realized) is that we need to do comparisons for angle rejection with the actual geometry's normal. Trying to do this after any underlying normal maps have been applied breaks the results.

We could probably avoid having `R3D_TARGET_GEOM_NORM_TAN` by using one of the existing textures (like the SSIL ones) to temporarily hold this data but having the raw geometry's normals/tangents seems useful enough to keep and can be useful at other stages, including post-processing.

One upside to this is there is no longer a combining pass for the normals since we can write directly to the regular g-buffer normal texture.

This is fully working but there are some things that are going to need to be addressed.

- The decal specific values are currently awkwardly held inside `r3d_draw_group_t` and a solution for where to store these needs to be found. The current idea is to remove the `R3D_Material` from `R3D_Decal` and hold all needed decal data directly in `R3D_Decal`. A lot of these are going to get assigned to the material in the draw call but there still needs to be a way to store non-material decal data somewhere in the draw call/group.

- The `visualize_to_screen` and `R3D_GetBufferNormal` functions are going to need to be adjusted because of how the normal/tanget render targets have changed.

- There are still unused uniforms that are being sent to the decal shader. This should be able to be dealt with with changes to scene.vert

- The decal example is cleaner now but there is still the possibility of it being replaced with a better one.

Also, I ran into rendering issues with `R3D_PRIMITIVE_CUBE` and it seemed the winding order of the back face triangles in `load_cube` were reversed. I just flipped the indices around but this could maybe use another look and changes to the actual vertex data. Either way it seems correct now.

I plan to review this soon and look at any feedback to hopefully get it completed to where it can be merged.

</details>


Bigfoot71:
Cool! I'll take a closer look at this later

Just a heads-up, I've made another PR that adds a linear depth buffer and tweaks quite a few things in the target management, it fixes quite a few problems that existed with screen space effects.

Just keep going like nothing happened, I'll sort out any conflicts _(it's going to be funny)_ if I can edit your branch.

Bigfoot71:
> We could probably avoid having `R3D_TARGET_GEOM_NORM_TAN` by using one of the existing textures (like the SSIL ones) to temporarily hold this data but having the raw geometry's normals/tangents seems useful enough to keep and can be useful at other stages, including post-processing.

For now keep it clear and simple, we can review the targets later, and why not add a temporary working buffer or something like that.

> * The decal specific values are currently awkwardly held inside `r3d_draw_group_t`

Revising the draw group might be complicated; either we keep it anyway and `r3d_draw_call_t` could be a union, or we create a separate list for the decals. I could handle that if necessary.

> * The `visualize_to_screen` and `R3D_GetBufferNormal` functions are going to need to be adjusted because of how the normal/tanget render targets have changed.

Noted, I'll take care of it when I sort out the conflicts with my new PR.

> * There are still unused uniforms that are being sent to the decal shader. This should be able to be dealt with with changes to scene.vert

AFAIK, unused uniforms will just be kept on the driver side so that it avoids crying error if we try to send something to it, but on the compiled code it will simply be ignored and will have no impact.

In any case, this is one of the classic drawback of uber shaders...

Furthermore, some data that is currently useless could become useful with custom shaders, so yeah, let's postpone that, but I'll keep it in mind, I'll see it anyway when adding the custom shaders support.

> * The decal example is cleaner now but there is still the possibility of it being replaced with a better one.

Cool, no need to do anything amazing, I think I'll make a more technical example repository once the first release is out.

> Also, I ran into rendering issues with `R3D_PRIMITIVE_CUBE` and it seemed the winding order of the back face triangles in `load_cube` were reversed. I just flipped the indices around but this could maybe use another look and changes to the actual vertex data.

Ah, quite possible, it was mainly used for cubemaps, and since face culling is disabled I may never have seen it...

grunthon:
The `R3D_Material` is out of `R3D_Decal` and all values are directly stored now. You can see how I added `R3D_GetBaseDecal` to create a decal without having uninitialized values. Let me know if there's any issues with this or any suggestions for a better way to handle it.

Bigfoot71:
> The `R3D_Material` is out of `R3D_Decal` and all values are directly stored now. You can see how I added `R3D_GetBaseDecal` to create a decal without having uninitialized values. Let me know if there's any issues with this or any suggestions for a better way to handle it.

If it's meant to work like `R3D_GetDefaultMaterial()`, the point is that we can override the default with a setter, and the instance is stored in `src/r3d_core_state.h`

That said, I'm not sure there's really a point for decals, I wouldn't have added any function otherwise.

At most, maybe a shortcut like `R3D_UnloadDecal()` to unload the material maps, but that can be thought about later if we also add a `R3D_LoadDecal()`.

It would be convenient, the functions for loading the material maps support sRGB.

---

Also, I just resolved the conflicts on my side, but I wasn't fast enough, let me know when you're done, and I should be able to push this to your branch.

grunthon:
> If it's meant to work like R3D_GetDefaultMaterial(), the point is that we can override the default with a setter, and the instance is stored in src/r3d_core_state.h

Yeah, it was really just a way to solve uninitialized "material" values that `R3D_GetDefaultMaterial` would have previously handled. I can look at coming up with something better, unless you want to take care of it.

> Also, I just resolved the conflicts on my side, but I wasn't fast enough, let me know when you're done, and I should be able to push this to your branch.

Sorry about that. I'm taking this out of draft. The main issue I still have is the storing of decal specific data inside of draw calls. It's certainly not good just having them in `r3d_draw_group_t`, and adding in a union somewhere seems like I'm just recreated the complexity that was nicely cleaned up by removing sprites, but I suppose there no way around doing something like it. If you're comfortable merging, go ahead. I don't want this sitting here building up more conflicts, especially when I haven't had time to look at your changes.

Bigfoot71:
Oh no no no, I just thought you were still making commits on your side

I didn't want to push to your branch if you had uncommitted work, it could have caused you trouble.
I'm going to push to your branch, but take your time, there's absolutely no problem.
I'm the one making changes, I'm the one fixing things x)

---

It's done, just remember to pull

Bigfoot71:
> It's certainly not good just having them in `r3d_draw_group_t`, and adding in a union somewhere seems like I'm just recreated the complexity that was nicely cleaned up by removing sprites

Regarding this point, the problem with sprites was that it is very easy to handle with what R3D provides.

For decals, calculating everything on the CPU would be a real pain, you'd really have to want to do it...
So yeah, I don't see a cleaner solution than using the union, the real problem is the unnecessary complexity.

But reusing existing structures is also an option, less clear, but acceptable.

As for the groups, thinking about it a bit more, it’s not really an issue.
Rendering individual meshes also consumes a single group per mesh, and culling by cluster works per group.
So if we exclude decals from the groups, we lose the clusters, which can relieve the CPU on large scenes, even though for decals, it remains a very specific case.

And maintaining a separate list with its own logic would surely add more complexity in the end.

Anyway, just tell me what you want to do, and what you'd prefer I do.
I was able to test the new example, and honestly, it's worth it, it's awesome!

include/r3d/r3d_material.h
```diff
@@ -22,9 +22,9 @@
 // ========================================
 
 /**
- * @brief Default environment configuration.
+ * @brief Default material configuration.
  *
- * Initializes an R3D_Environment structure with sensible default values for all
+ * Initializes an R3D_Material structure with sensible default values for all
```
Bigfoot71, line 27: hope this doesn't show that I'm a bit lazy x)

shaders/scene/decal.frag
```diff
@@ -11,75 +11,108 @@
...
-    // Sample albedo texture and discard if below alpha cutoff
+    /* Compute albedo and discard if below alpha cutoff */
     vec4 albedo = vColor * texture(uAlbedoMap, decalTexCoord);
     if (albedo.a < uAlphaCutoff) discard;
 
-	// Apply material values
-    FragAlbedo = albedo;	
-    FragEmission = vec4(vEmission, 1.0) * texture(uEmissionMap, decalTexCoord);
+    /* Retrieve surface normal */
+    vec3 surfaceNormal = M_DecodeOctahedral(texture(uNormTanTex, fragTexCoord).rg);
```
Bigfoot71: There is `V_GetWorldNormal(tex, coord)` function for this, which might be clearer to read and safer if we decide to change which space is stored, there's even a version that does it using texelFetch now.

include/r3d/r3d_material.h
grunthon: Hey, I figure if I can fix some typos and copy-paste screw ups it might even out for the ones I get caught making.

src/modules/r3d_shader.h
```diff
@@ -580,6 +582,7 @@ typedef struct {
 
 typedef struct {
     unsigned int id;
+    r3d_shader_uniform_sampler_t uBoneMatricesTex;
```
Bigfoot71: Actually, it wasn't strictly necessary to get the unused uniforms in this context. If we had tried to define them, it simply wouldn't have compiled.

I'm just leaving it as a note, all of this will be reviewed for the custom shaders in the next version.

shaders/scene/decal.frag
Bigfoot71: Edit: I doubt we'd want to change the space for `uNormTanTex`, but oh well, it's been said!

grunthon:
Regarding `R3D_GetBaseDecal`, I'm not sure of a much better way to handle all the uninitialized values when creating an `R3D_Decal`. Without something like this it's going to be up to the user to initialize all the values, which is an option but not a very user friendly one. There could be a macro instead that would be used like `R3D_Decal d = R3D_DECAL_BASE`, but that's really not much different though possibly clearer in how it functions.

grunthon:
> Anyway, just tell me what you want to do, and what you'd prefer I do.

As for the decal specific data, draw calls and groups, I've wound up going in circles a bit on it. I suppose if I were to decide I would say just having a union in `r3d_draw_call_t` for geometry and decals is what I would choose. I don't really like it adding complexity to everything else related to draw calls, but I'm not really coming up with anything better.

There could be specifics to this that you would be more aware of by being more familiar with the whole code base, so I'll defer to any solution you might prefer. Or just let me know if you'd rather handle it on your own.

Bigfoot71:
> There could be a macro instead that would be used like `R3D_Decal d = R3D_DECAL_BASE`, but that's really not much different though possibly clearer in how it functions.

Having it as a constant only would be more consistent with the rest than having it as a function, but to avoid an empty header, a function to unload the maps as I mentioned would be a small comfort...

We'll see later if there's more of a need, but it seems pretty good to me.

> As for the decal specific data, draw calls and groups, I've wound up going in circles a bit on it. I suppose if I were to decide I would say just having a union in `r3d_draw_call_t` for geometry and decals is what I would choose.

Clearly, decal related data shouldn't be in `r3d_draw_group_t`.
But looking at it more closely, I'm not sure it actually makes things more complex.

Tagged unions aren't the most elegant thing in C imo, but aside from that, we already have the draw call push function that takes a bool to indicate whether it's a decal or not, which already implies "special-case" logic for decals:
```c
void r3d_draw_call_push(const r3d_draw_call_t* call, bool decal)
```

So even with a union, the push function would stay largely the same, just with one less parameter. After that, it’s just a matter of iterating through the decal list, and that's it, `raster_decal()` in `src/r3d_draw.c` already handles everything.

I don't get the impression that this adds more complexity, like a bunch of switches or other annoying things.
In the end, it shouldn't make much difference.

If you want, I can make this change directly on your branch if you're not sure, just let me know.

grunthon:
I have removed `R3D_GetBaseDecal` and `R3D_DECAL_BASE` can be used directly now instead.

`R3D_UnloadDecalMaps` has been added which just calls the other map unloading functions. The only issue I could see is if it is called using a decal that had uninitialized texture ids for any of the maps, but in that case the decal was not correctly initialized anyway so that's really a different problem.

> If you want, I can make this change directly on your branch if you're not sure, just let me know.

I'll just let you make the changes you see fit. Though I don't mean to make more work for you, it's probably better than going back and forth about it anymore.

Beyond this, I don't believe there is really anything more to be dealt with that isn't beyond the intended scope of the PR. Let me know what you think.

_(Just to note, the naming of anything added in this is open to change to be more suitable or in line with the library as a whole. Feel free to make any changes or let me know if I can adjust anything.)_

Bigfoot71:
Okay, I've implemented the tagged union for `r3d_draw_call_t`, like this:
```c
typedef enum {
    R3D_DRAW_CALL_MESH,
    R3D_DRAW_CALL_DECAL
} r3d_draw_call_enum_t;

typedef struct {
    R3D_Material material;
    R3D_Mesh instance;
} r3d_draw_call_mesh_t;

typedef struct {
    R3D_Decal instance;
} r3d_draw_call_decal_t;

typedef struct {
    r3d_draw_call_enum_t type;
    union {
        r3d_draw_call_mesh_t mesh;
        r3d_draw_call_decal_t decal;
    };
} r3d_draw_call_t;
```

Nothing related to decals is stored in `r3d_draw_group_t` anymore.

And I also made quite a few adjustments along the way.
Notably, I removed `modules/r3d_primitive.c` and put all that logic directly into `modules/r3d_draw.c`.

This way, everything that's supposed to lead to draw calls is centralized, and `r3d_draw()` and `r3d_draw_instanced()` handle everything automatically depending on the type of draw call provided.

_(All of this could have been deferred to another PR, but at least we can merge in a clean way)_

So, apart from all the adjustments, I honestly don't find it any more complex than before.
Let me know if you were planning on making any changes; otherwise, I can merge it!

Bigfoot71:
> I don't mean to make more work for you

Honestly, I don't care, if it has to be done I'll do it with pleasure!
_(if it bothers me, I'll say so without holding back)_

> The only issue I could see is if it is called using a decal that had uninitialized texture ids for any of the maps

Yeah, it should be safe to pass something null to a free function, I'll look into that after this PR.

> Beyond this, I don't believe there is really anything more to be dealt with that isn't beyond the intended scope of the PR.

Yeah, apart from these last adjustments, the expected result is there, if you see nothing more to do you tell me.

> (Just to note, the naming of anything added in this is open to change to be more suitable or in line with the library as a whole. Feel free to make any changes or let me know if I can adjust anything.)

And for that, there's no more inconsistency, it's all good!

src/modules/r3d_shader.h
```diff
@@ -840,6 +846,7 @@ static const struct r3d_shader_loader {
         r3d_shader_loader_func ambient;
         r3d_shader_loader_func lighting;
         r3d_shader_loader_func compose;
+        r3d_shader_loader_func composeNormals;
```
Bigfoot71: The `composeNormals` shader is still here, but it doesn't seem necessary anymore?

src/modules/r3d_target.h
```diff
@@ -25,11 +25,12 @@
 typedef enum {
     R3D_TARGET_INVALID = -1,
     R3D_TARGET_ALBEDO,          //< Full - Mip 2 - RGB[8|8|8]
-    R3D_TARGET_NORM_TAN,        //< Full - Mip 2 - RGBA[16|16|16|16]
+    R3D_TARGET_NORMAL,          //< Full - Mip 2 - RGBA[16|16]
```
Bigfoot71: Typo `RGBA[16|16]`, but I think it's my fault, when I resolved the conflicts with master

src/r3d_utils.c
```diff
@@ -52,7 +52,7 @@ Texture2D R3D_GetNormalTexture(void)
 Texture2D R3D_GetBufferNormal(void)
 {
     Texture2D texture = { 0 };
-    texture.id = r3d_target_get(R3D_TARGET_NORM_TAN);
+    texture.id = r3d_target_get(R3D_TARGET_GEOM_NORM_TAN);
```
Bigfoot71: It should be `R3D_TARGET_NORMAL` here, and for the format, since raylib doesn't have `R16G16`, we can use `R32`, it's good enough x)

I could review this after the PR though.

src/modules/r3d_shader.h
grunthon: Yeah, I just missed it. I'm going to take a quick look over this shortly and correct this and the other issues mentioned in your reviews.

grunthon:
Your changes are good and I don't see any other issues so feel free to merge at your discretion.

Bigfoot71:
Thank you for your time and contribution!

grunthon:
I've edited the initial comment (while retaining the original for reference) to more concisely reflect the final changes.

PR #150, opened 2026-01-15
Review normal buffers formats

Bigfoot71:
@grunthon I'm making a PR so it's easier to keep you informed.

I wanted to review the render targets to see how we can save memory, and I thought normals could be stored as unsigned short, it doesn't save anything, but it's more logical and could be _slightly_ cheaper on some hardware.

Then I realized the `GEOM_NORM_TAN` target isn't meant to store detailed info, just the surface info without bump mapping, so I downgraded it to 8-bit unsigned to see.

From experience, octahedron encoding with 8-bit precision can cause "banding" with reflections, like described here: https://knarkowicz.wordpress.com/2014/04/16/octahedron-normal-vector-encoding/

But for decals the precision seems sufficient at first glance, I would like to have your opinion, if you have anything to say about it before merging.

_ps: I shrunk the images, but you can open them bigger, though I honestly see no difference… maybe I'm just blind x')_

---

### 16-bit precision:
screenshot001

### 8-bit precision:
screenshot000

---

### 16-bit precision
screenshot001

### 8-bit precision
screenshot000


grunthon:
I did a quick test and also found no noticeable change in quality. Out of curiosity I did an image difference between two screenshots of 8 and 16 bit which only showed extremely minor differences. So I don't see any problem doing it, and it's not something that's difficult to change later.

I didn't bring this up in #147, but I had already thought there could be a better way to store the raw surface normal/tangent data. Since the tangent should always be perpendicular to the normal (some lighting / effects could end up with them out of alignment, but not in this case) it seems like it should be possible to store the normal at high precision while storing the tangent as only its rotation around the normal. I haven't really thought it through but the idea is relevant to this.

Bigfoot71:
Hehe, super smart!

I tried to implement your idea and break down the problem like this

To encode the angle:
- First, I get a reference vector `up` chosen to avoid being aligned with the normal
- Then, from this reference, I construct an orthonormal basis `(refT, refB)` perpendicular to the normal, which is therefore deterministic
- Finally, I measure the angle of the tangent with respect to `refT` using `atan()`

Then, in the decal shader, simply:
- Reconstruct the orthonormal basis `(refT, refB)` from the normal (deterministic)
- Rotate `refT` by the recorded angle to find the tangent

Here are the functions:
https://github.com/Bigfoot71/r3d/blob/a6cda0827dac3ae4df2f4f1c30641ed14299e895/shaders/include/math.glsl#L70-L87

Perhaps I've overthought the problem, let me know if you had something else in mind.

But for now, everything is done in RGB8, and it seems okay so far.
I'm going to try some more complex scenarios to see if it works well.

After all, if the flaws are only noticeable when you know they're there and you squint at it for more than five seconds, I'd say it's a success x)

Small sphere/cylinder captures:

screenshot000

screenshot000

grunthon:
Nice. Yeah that's generally what I was thinking. As for 8/16 bit I did a bit more testing and when using angle cutoff and fading I am not getting good results when using 8 bit values, as shown below. I had wondered if there is some clever way to encode normal + tangent angle into just RG16F, but I'd have to think about it or do some research.

GL_RGBA8
nt8

GL_RGBA16F
nt16

Bigfoot71:
Yeah, I tried it and yes, it does show those little squares...

As for the idea of ​​storing everything in a vec2, I don't see an elegant solution, and I have a feeling that if there is one, it would be really crazy

Or we don't store the tangent, and calculate it from the UVs, I tinkered with this:
```glsl
vec2 dUVdx = dFdx(decalTexCoord);
vec2 dUVdy = dFdy(decalTexCoord);
vec3 dPdx = dFdx(positionViewSpace);
vec3 dPdy = dFdy(positionViewSpace);

vec3 T = normalize(dPdx * dUVdy.y - dPdy * dUVdx.y);
vec3 B = normalize(dPdy * dUVdx.x - dPdx * dUVdy.x);

T = normalize(T - surfaceNormal * dot(surfaceNormal, T));
B = normalize(cross(surfaceNormal, T));
```

It seems good so far, but more testing is needed to be sure.

grunthon:
> As for the idea of ​​storing everything in a vec2, I don't see an elegant solution, and I have a feeling that if there is one, it would be really crazy

Yeah you're probably right. I did think about it a bit more and didn't come up with anything that made a lot of sense.

> Or we don't store the tangent, and calculate it from the UVs

Ah yes that's interesting. It's similar to how I was originally trying to derive the normal/tangent but just for the tangent. I might play around with the idea as well. Visualizing how it compares to the actual surface tangent could be useful.

I do suspect that just storing the normal and tangent angle is probably the best overall solution. If the main goal here is to reduce memory usage of the render targets, I had previously mentioned we can just temporarily use one of the other textures, like `R3D_TARGET_SPECULAR`, instead of even having `GEOM_NORM_TAN`. At least If we're only thinking of their use in decals. Of course that won't reduce actual memory bandwidth usage.

Bigfoot71:
> I had previously mentioned we can just temporarily use one of the other textures, like `R3D_TARGET_SPECULAR`

Yes, the only annoying thing is the calirty, it's better to be clear than unclear just to save 10mb.

I'll see how to do it, the idea would be to have a temporary buffer somewhere that can be reused several times in an acceptable way.

Because the format matter too, if the format is too large for what we are doing at a moment, it consumes more bandwidth, so it is indeed an option but something to think about.

Regarding the tangent angle, I'll compare whether it saves any processing time. If it does, I'll leave it as RGB16U, otherwise, I'll revert it to RGBA16U.

And because just that, as it stands, we can't say it saves memory, something like RGB16F may be allocated/aligned in VRAM as if it were RGBA16F for efficiency reasons, and therefore may not actually save any memory.

That might not be the case, but we have no control over that, and the spec obviously doesn't require anything in that regard.

grunthon:
It seems like you have enough to go on to decide what you want to do. Just let me know if there's anything else I can do to help.

Bigfoot71:
Okay, regarding efficiency, octahedron encoding requires fewer operations on the ALU, and since it doesn't necessarily save memory, sticking with the RGBA approach seems better.

However, the method where tangents are obtained from UV and view position derivatives seems better, in that it requires fewer operations and saves a bit of bandwidth, and I don't find any obvious artifacts, I just had to adjust the previous method to account for the sign of the bitangent.

So, in the end, I think we can just store the surface normal.

I even thought we could apply a write mask to the normal output to store both normals in a single buffer, but that's starting to feel like a tricky workaround, and it seems that it requires `GL_ARB_draw_buffers_blend`, which is an extension in GL 3.3

Bigfoot71:
Otherwise, small adjustment, I vectorized the condition here:
```diff
/* Discard fragments that are outside the bounds of the projector */
- if (abs(positionObjectSpace.x) > 0.5 ||  abs(positionObjectSpace.y) > 0.5 || abs(positionObjectSpace.z) > 0.5) discard;
+ if (any(greaterThan(abs(positionObjectSpace.xyz), vec3(0.5)))) discard;
```

Also, I saw that for the emission output, you put `fadeAlpha` in `.w`, what was that for?

https://github.com/Bigfoot71/r3d/blob/624b5530cddbc7e760b47b42540945a6011cfe0c/shaders/scene/decal.frag#L115-L117

---

And finally, I noticed that sometimes the shadow at the "bump" was inverted in the decal depending on the face of the cube where it is projected, it may come from the rotations in the example, because it also happens on the master branch, I will look into it.

grunthon:
> Also, I saw that for the emission output, you put fadeAlpha in .w, what was that for?

The intent was to correctly fade the emission output the same as everything else.

> And finally, I noticed that sometimes the shadow at the "bump" was inverted in the decal depending on the face of the cube where it is projected, it may come from the rotations in the example, because it also happens on the master branch

Hmm. This is something I had encountered while working on it but thought it had been corrected. I will check on it as well when I have some time soon.

Bigfoot71:
> The intent was to correctly fade the emission output the same as everything else.

I asked because blending is enabled when rendering decals, but it's not redefined on a specific function and I think that would be incorrect for normals, it would be better to have blending per buffer, so we're back to `GL_ARB_draw_buffers_blend` :/

Or we could force '.a = 1.0' for normals, but that seems strange to me, maybe play with the normal scale too, I'll think about it.

grunthon:
> Or we could force '.a = 1.0' for normals, but that seems strange to me, maybe play with the normal scale too

The combination of both is actually how it currently functions.

https://github.com/Bigfoot71/r3d/blob/00a7710c026eb860fabaebc40bc2a2d3f999fdaa/shaders/scene/decal.frag#L102-L103

Bigfoot71:
Ah, I reacted for nothing, I could've reread it in the meantime x)
Sorry, I'll run more tests in a bit and see if we merge it like that

grunthon:
> I noticed that sometimes the shadow at the "bump" was inverted in the decal depending on the face of the cube where it is projected, it may come from the rotations in the example

I did look into this and yeah it does seem to be from the rotation of the cube faces being used for the rotation of the decals. I think I might go ahead and make a simpler decal example that just shows a few decals being drawn. If you like the current one and would prefer to just get it fixed or make a new one yourself, let me know.

Bigfoot71:
> If you like the current one and would prefer to just get it fixed or make a new one yourself, let me know.

If you're up for it, I'll merge it!

As mentioned before, I plan to create a repo with more advanced examples after the "final" release, so simplifying the current example is fine, it'll likely just serve for quick tests.

_But if you don’t feel like bothering with it, it's not critical either, in my opinion_

grunthon:
I'll just make a new really basic example soon. It's no problem.

I did try your code for calculating the tangent by including the UVs and was getting a lot of artifacts and unintended rotation of the tangent. I presume at least partly because the view space transformation isn't being included. I had these kind of problems a lot when I was previously trying to derive the normal and tangent.

dtan

Bigfoot71:
> I presume at least partly because the view space transformation isn't being included. I had these kind of problems a lot when I was previously trying to derive the normal and tangent.

Ouch, that's exactly what I was afraid of, which is why I didn't merge it.
If you have any other ideas, feel free to share them, I'll think about it as well.

But if nothing conclusive comes up, I think encoding the tangents with normals in RGBA would remain acceptable for 0.7

The angle idea didn't provide enough clear advantages to be worth keeping, in my opinion, at least not in the form I had implemented it.

Bigfoot71:
I'm closing this PR, we'll keep the RGBA buffer with full normals + tangents encoded.
I'll open another one to incorporate the minor adjustments I've made.

PR #153, opened 2026-01-16
Fix syntax when clearing SSIL render target

grunthon:
Was missing the depth argument for `R3D_TARGET_CLEAR`.


Bigfoot71:
Oops, thanks for fixing that!

PR #159, opened 2026-01-18
Correctly handle null albedo texture for decals

grunthon:
The main purpose of this is to correctly render normal-only decals when no albedo texture is set. There are other ways this could be handled but this keeps it simple and just disables the output when the texture is null.

### DIFF
--- a/examples/decal.c
+++ b/examples/decal.c
@@ -28,7 +28,7 @@ int main(void)
 
     // Create decal
     R3D_Decal decal = R3D_DECAL_BASE;
-    decal.albedo = R3D_LoadAlbedoMap(RESOURCES_PATH "images/decal.png", WHITE);
+    //decal.albedo = R3D_LoadAlbedoMap(RESOURCES_PATH "images/decal.png", WHITE);
     decal.normal = R3D_LoadNormalMap(RESOURCES_PATH "images/decal_normal.png", 1.0f);
     decal.normalThreshold = 89.0f;
 
--- a/shaders/scene/decal.frag
+++ b/shaders/scene/decal.frag
@@ -42,6 +42,7 @@ uniform vec2 uTexCoordScale;
 
 uniform float uNormalThreshold;
 uniform float uFadeWidth;
+uniform bool uAlbedoEnabled;
 
 /* === Fragments === */
 
@@ -103,7 +104,7 @@ void main()
     FragNormal = vec4(M_EncodeOctahedral(N), 0.0, 1.0);
 
 	/* Apply material */
-    FragAlbedo = albedo;
+    if (uAlbedoEnabled) FragAlbedo = albedo;
     FragEmission = vec4(vEmission * texture(uEmissionMap, decalTexCoord).rgb, fadeAlpha);
 
     vec3 orm = texture(uOrmMap, decalTexCoord).rgb;
--- a/shaders/scene/decal.frag
+++ b/shaders/scene/decal.frag
@@ -105,6 +105,9 @@ void main()
     if (uAlbedoEnabled) {
         FragAlbedo = vec4(albedo.rgb, albedo.a * fadeAlpha);
     }
+    else {
+        FragAlbedo = vec4(0);
+    }
     
     FragEmission = vec4(vEmission * texture(uEmissionMap, decalTexCoord).rgb, fadeAlpha);
 
--- a/shaders/scene/decal.frag
+++ b/shaders/scene/decal.frag
@@ -83,11 +83,9 @@ void main()
     /* Discard if the angle exceeds the threshold */
     if (difference < 0.0) discard;
 
-    /* Fade on edge of threshold */
-    // note: could do another (or the main) uAlphaCutoff check here for discard
+    /* Compute fade on edge of threshold */
     float fadeAlpha = clamp(difference / uFadeWidth, 0.0, 1.0);
-    albedo.a *= fadeAlpha;
-
+ 
     /* Retrieve surface tangent */
     vec3 surfaceTangent = M_DecodeOctahedral(texelFetch(uNormTanTex, ivec2(gl_FragCoord.xy), 0).ba);
 
@@ -104,7 +102,10 @@ void main()
     FragNormal = vec4(M_EncodeOctahedral(N), 0.0, 1.0);
 
 	/* Apply material */
-    if (uAlbedoEnabled) FragAlbedo = albedo;
+    if (uAlbedoEnabled) {
+        FragAlbedo = vec4(albedo.rgb, albedo.a * fadeAlpha);
+    }
+    
     FragEmission = vec4(vEmission * texture(uEmissionMap, decalTexCoord).rgb, fadeAlpha);
 
     vec3 orm = texture(uOrmMap, decalTexCoord).rgb;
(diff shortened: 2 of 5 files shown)


Bigfoot71:
Normally, a null texture should fallback to a default texture like white, black, or blank. Not writing to an active output is undefined behavior.

_Or rather, the `out` value could have any value because it is uninitialized and be output as such._

This is already the case, the default texture is "WHITE":
https://github.com/Bigfoot71/r3d/blob/329eb663e52de28a3d1c4b0d1013851ae813c990/src/r3d_draw.c#L958

Perhaps replacing "WHITE" with "BLANK" would be the solution, if I understand the idea correctly.

grunthon:
> Not writing to an active output is undefined behavior.

Hm yes, I did not consider that.

> Perhaps replacing "WHITE" with "BLANK" would be the solution

This issue with that is the shader will always discard since the alpha is below `uAlphaCutoff`. I have instead added a zero output for when the albedo is disabled.

I'm certainly open to any better way to this. The goal is correctly draw normals (and other material values) without any change to albedo. Useful for things like footsteps in snow, bullet holes in walls, etc.

Bigfoot71:
~~Instead of discarding, could we use a mask for out of bounds fragments that sets fadeAlpha to zero?~~

Sorry, I forgot about the alpha cutoff, but is it really useful since we blend anyway?
I mean, without sorting, for the bounds, yes, but what about the transparency itself?

I'm working on something else right now, but I could test it later, I'm only talking off the top of my head...

grunthon:
Yeah, I also realized the the alpha check shouldn't really even be happening if we're not using it. (That's what I get for messing with this right after waking up)

I do think some kind of mask to limit the area the normal is drawn (when not using an albedo texture) could be a good idea since we're currently completely overwriting the normal. I was initially going put a user settable flag in `R3D_Decal` to disable albedo output. Maybe having something like that but still being able to use the alpha from an albedo texture as a mask could work?

I'll think about this some more.

grunthon:
I added a `disableAlebdoColor` flag to `R3D_Decal` along with a corresponding uniform in the shader that results in `FragAlbedo` being set to `vec4(0.0)` when enabled. The uniform is set true if either the albedo texture is null or `disableAlebdoColor` is set. This handles either not using an albedo texture as well as only using the albedo alpha as a mask.

For normals this only works through the alpha check for discard as we can't blend the normals since we're directly writing to the g-buffer normal texture. Blending normals would require going back to having a combining pass for normals and more render target juggling that I wasn't planning on getting into. At least not right now.

I wanted to see what you thought before making a commit.

Bigfoot71:
Yeah, I've been able to do some testing on my end and I understand the necessity of all this now.
I think it's best what you've done, meaning a boolean to indicate whether there's a albedo to sample, that's perfect.

We can just tweak a few minor details, like avoiding binding the texture if it's unnecessary, but that's just nitpicking.
The current state seems best, so I'm merging it!

_I'm merging this directly because I'm going to close the other PR and add the small adjustments I made there._
_I'll ping it back here afterward_

grunthon:
Just to be clear, the changes mentioned in the previous comment had not been committed yet. Did you mean to merge it as it is without the proposed addition of `disableAlebdoColor` in `R3D_Decal`? I meant to mark this as a draft.

I also had loading of the albedo texture commented out in the example, which was of course not meant to be included.

Bigfoot71:
> I also had loading of the albedo texture commented out in the example, which was of course not meant to be included.

Yeah, no problem, I had noticed it and fixed it here: https://github.com/Bigfoot71/r3d/commit/10fd60de11351300d83f8feb8328ef07b9d1420a

> Did you mean to merge it as it is without the proposed addition of `disableAlebdoColor` in `R3D_Decal`?

For that, there's actually no need for an extra flag, the absence of an albedo should be enough to indicate that we don't want one.

If for some reason someone wants a white texture, they can get the internal one or the one from raylib, so that's fine.

---

I made a few adjustments here: https://github.com/Bigfoot71/r3d/commit/6efebbaf46b74f09aaf95a4fff21f96df9190d4f

_Ignore the comment about the conditional albedo bind, I've fixed it: https://github.com/Bigfoot71/r3d/commit/4ff0c94d639348c7c87845723ce6b438873d3ad5_

But yeah I merged too fast, I'm hoping to stabilize everything a bit for 0.7 and finally start the custom shaders, but if the decals will still require work we can wait.

For normals, it's clearly a bit of a problem, the current state makes rendering without albedo not very useful, and if we solve this normal issue with a suitable blending, we should be able to get away with a simple blank texture for albedo in the end, I imagine, _or maybe not, idk, just a thought x)_

Well, no problem, it's merge, not ideal but acceptable.

Let me know if you plan to review all of this, and how. If so, but not immediately, that's not a problem at all, but the final state of how the decals are made will surely influence the custom shaders and I'd like them to be compatible with the decals as well.

grunthon:
No problem. Sorry for the confusion. I've been jumping between things and having "one of those days".

The main issue with not having some sort of mask for drawing normals (which having the extra flag allowed) is the problem of overwriting the underlying normal mapping in an undesirable way.
normals

The alternative is, as you say, coming up with a solution for actually blending the normals. But even then having a mask could still be useful.

I do believe there could be one other problem with the decal normals. It seems since the rotation of the decal isn't accounted for in the TBN the result can be incorrect. At least that's what I think is happening. It could be another reason why it sometimes seems "inverted" in the current example.

I don't want any of this to slow overall progress. Obviously the decal normal mapping has turned out more complicated than I originally thought. If there's anything you need to do with the decals to keep things moving forward I hope you don't feel that you have to check with me.

Bigfoot71:
I ran a few tests and I can see how normal accumulation could be implemented, but it would indeed require adding yet another buffer. At that point it starts to feel like too much for this context. In a different setup I'd probably be fine with it, but for a raylib / GL 3.3 renderer it feels overkill.

A more pragmatic solution would indeed be to rely on a mask texture, or to repurpose `uAlbedoEnabled` into something like `uApplyColor`. When color application is disabled, we would still use the alpha channel for cutoff, but at the final output stage we simply blend the color to zero so it doesn't affect the result.

The main drawback is that if we only need a mask, we're still reading from the `.a` component, effectively wasting three channels. Swizzling could mitigate this, but raylib doesn't really expose that directly.

Adding a dedicated mask texture is another option, but it raises questions about how the albedo alpha should be interpreted. Ignoring it entirely could make the shader logic a bit messy imo.

Given all that, I think the most pragmatic approach here is to introduce a `uApplyColor` parameter, keep using alpha as a mask, and simply blend the color to zero when it's disabled.

It's not a perfect solution, but in this context no solution really is.

And if someone is extremely concerned about "wasting" three texture channels (even though the actual gain would be marginal), they can still rely on swizzling on their side.

I'll open a separate PR for this. Feel free to review it before merge if you'd like.

grunthon:
To briefly bring this back up, do you feel there isn't a need for a null albedo texture check when setting `uApplyColor`? I can understand not wanting to add another check but creating a decal with only non-albedo textures loaded does result in the `WHITE` texture being visible.

noalbedo

Of course setting `applyColor` handles this, but that might not be obvious and possibly unintuitive.

I'm pretty much done with decals for the time being unless something comes up. This was really the only thing I thought might want to be addressed before a new release.

Bigfoot71:
Yeah, we could say that if there's no albedo, then the behavior is "no color, mask/alpha at 1.0" which would be done by internally forcing `applyColor` to false, but still binding a texture with an alpha of 1.0

That would seem more logical, you're right. I could make the change tomorrow.

Bigfoot71:
I made the adjustment here: https://github.com/Bigfoot71/r3d/commit/c24245cc9d46da9d0bb9c81e389a7571e42723a8

PR #160, opened 2026-01-20
Decal mask

Bigfoot71:
Adds a boolean `applyColor` to `R3D_Decal` so that only the albedo alpha is used as a mask.

See the comment here for more details: https://github.com/Bigfoot71/r3d/pull/159#issuecomment-3773291359

The decal shader was also slightly reorganized by the way.

I tried to group texture sampling as much as possible, revisited the handling of handedness, and lightened the comments a bit since the code is fairly clear.

@grunthon I'll give you veto power if there's any issue!

Here is an example with `applyColor` set to false:

[Enregistrement d'écran_20260120_162119.webm](https://github.com/user-attachments/assets/d645cd22-7c7b-4951-8c7f-51339dc822be)

### DIFF
--- a/include/r3d/r3d_decal.h
+++ b/include/r3d/r3d_decal.h
@@ -51,7 +51,8 @@
         .uvScale = {1.0f, 1.0f},                        \
         .alphaCutoff = 0.01f,                           \
         .normalThreshold = 0,                           \
-        .fadeWidth = 0                                  \
+        .fadeWidth = 0,                                 \
+        .applyColor = true                              \
     }
 
 // ========================================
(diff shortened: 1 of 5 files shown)


Bigfoot71:
Oh and I also added `fadeAlpha` for the ORM, which wasn't there before.
It's better than nothing I guess, it could have caused some issues otherwise.

grunthon:
Looks good to me. I was going to mention the ORM. I had added fading for it but didn't make a commit for it #159.

It is better to sample the normal-tangent texture together like that, and that's a nice way to deal with the handedness check/correction as well.

And yeah it was too comment heavy as a byproduct from it being worked on.

Bigfoot71:
> And yeah it was too comment heavy as a byproduct from it being worked on.

No problem, even I comment a lot sometimes, like it's done in raylib, it can serve as a learning reference for others.

---

Okay, so I'll merge!

I'm currently looking into the orientation issue, and I think I actually made a mistake last time with tangents from derivatives, which caused that weird noise/band effect.

I'm now basing on this article: http://www.thetenthplanet.de/archives/1180

I've gone through everything again and I no longer see those strange patterns.

If this works, we won't need tangents in the buffer anymore, but there's still shadow inversion, different this time, I'm trying to fully understand what's going on.

grunthon:
I'll have to look at that article when I have a chance. I had not come across that one.

The specific rotation issue I recently mentioned is related to the rotation of the decal bounding box and not the tangent of the surface we are drawing on. Accounting for the model transformation would seem to be what's needed, but I haven't had time to look into it more yet.

https://github.com/user-attachments/assets/ecfc168c-4e34-4e4a-a0f8-01dc12e6aa15

PR #161, opened 2026-01-20
Decal tangent fix

Bigfoot71:
@grunthon I was able to fix all the apparent issues. Hopefully this doesn't introduce new ones, but there's no reason it should.

The first thing I revised was to stop relying on the surface tangent, and instead rebuild the TBN in decal space:
```glsl
mat3 BuildDecalTBN(vec3 worldNormal)
{
    vec3 decalX = normalize(vDecalAxes[0]);
    vec3 decalZ = normalize(vDecalAxes[2]);

    vec3 T = normalize(decalX - dot(decalX, worldNormal) * worldNormal);
    vec3 B = normalize(decalZ - dot(decalZ, worldNormal) * worldNormal);

    B *= sign(dot(cross(T, B), worldNormal));

    return mat3(T, B, worldNormal);
}
```

This approach projects the decal's X and Z axes onto the surface's tangent plane, ensuring that bump mapping normals rotate with the decal rather than with the surface geometry.

Next, I also simplified the varyings:
```glsl
#if defined(DECAL)
    vDecalProjection = inverse(finalMatModel) * uView.invView;
    vDecalAxes = mat3(finalMatModel);
#endif // DECAL
```

`vDecalAxes[1]` now directly provides the correct projection axis for the angle calculation in the fragment shader, eliminating the need for a separate orientation varying.

Everything works correctly now. Here's a demo, these are supposed to be hollows _(decal rotation at the end)_:

[Enregistrement d'écran_20260120_213235.webm](https://github.com/user-attachments/assets/986dd934-ce3b-454e-99b5-5a4dbea5299d)

But the shadows were still inverted with the `decal_normal.png` texture because the Y channel was inverted (DirectX convention instead of OpenGL). I fixed it with Gimp.

[Enregistrement d'écran_20260120_214942.webm](https://github.com/user-attachments/assets/29cd2416-e0d9-49a6-8081-5d51150081e4)

Everything appears to be working now.


grunthon:
Ah very nice. I will test this out soon.

A comment I recently made on the last PR actually had me thinking that for the tangent maybe we should be starting from the decal's local space instead of trying to correct for it.

I think I had actually inverted the normal texture with the intent of making it correct for OpenGL when apparently it already was. Ugh.

grunthon:
https://github.com/Bigfoot71/r3d/blob/cf6542186e72ca43886f1d0f03474f20309a26bb/shaders/scene/decal.frag#L93
I believe the rotation vector needs to be normalized here to work correctly.
`float angle = acos(clamp(dot(normalize(vDecalAxes[1]), worldNormal), -1.0, 1.0));`

Bigfoot71:
> I believe the rotation vector needs to be normalized here to work correctly.

Ah yes thank you, you're right, I normalized the axes in the vertex shader.

Bigfoot71:
I was able to test a few other scenarios without any apparent problems, so I'm merging.

grunthon:
Thanks for the assistance with this.

PR #162, opened 2026-01-21
Add Ex and Pro decal drawing functions

grunthon:
The decal drawing functions are now generally in line with the mesh drawing functions.

- `R3D_DrawDecal(R3D_Decal decal, Vector3 position, float scale)`
- `R3D_DrawDecalEx(R3D_Decal decal, Vector3 position, Quaternion rotation, Vector3 scale)`
- `R3D_DrawDecalPro(R3D_Decal decal, Matrix transform)`
- `R3D_DrawDecalInstanced(R3D_Decal decal, R3D_InstanceBuffer instances, int count)`
- `R3D_DrawDecalInstancedEx(R3D_Decal decal, R3D_InstanceBuffer instances, int count, Matrix transform)`

I also adjusted the comment blocks of some the other draw functions to be a bit more self-consistent.

### DIFF
--- a/examples/decal.c
+++ b/examples/decal.c
@@ -118,7 +118,7 @@ int main(void)
             R3D_DrawDecalInstanced(decal, instances, decalCount);
         }
 
-        R3D_DrawDecal(decal, MatrixTransform(hitPoint, decalRotation, decalScale));
+        R3D_DrawDecalEx(decal, hitPoint, decalRotation, decalScale);
 
         R3D_End();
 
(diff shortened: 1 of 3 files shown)


Bigfoot71:
Let's goooo! Thank you for that!

I'm going to do a quick review/cleanup of the project, and I think we're good to go for 0.7!
I'll start working on the custom shaders directly on a separate branch.

grunthon:
I'm going to make a new decal example in just a bit and do a little more testing. After that decals should be in a pretty good state.

PR #164, opened 2026-01-21
New decal example

grunthon:
This should be just enough to show creating a decal and using the draw functions. Using instancing probably isn't really needed, but eh. Let me know if it seems alright.

example


Bigfoot71:
Well, that's great! We can see almost everything in just under 100 lines, so it's perfect!
Just asking, if you don't want to do any editing, otherwise I'll press the button!

grunthon:
Should be good to go.

Also, I didn't run into any issues when testing or working on this so decals seem stable for now.

Issue #156, opened 2026-01-18
PBR example: IMAGE: Data format not supported warnings on Windows (MSVC)


sacplex:
When loading the PBR example in r3d, I consistently receive multiple image loading warnings.

This happens on Windows 11 using MSVC.

INFO: R3D: Loading textures with 8 worker threads
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
WARNING: IMAGE: Data format not supported
WARNING: IMAGE: Failed to load image data
INFO: R3D: Loaded 8 textures successfully

grunthon:
Is this using the current master branch or one of the releases? Are you using the included version of Raylib?

There were recent changes to the Raylib build configuration in the master branch so if you have updated from it you might want to make sure your build environment is up to date by rerunning CMake then doing a clean build.

If you are use your own build of Raylib you're going to want to have support for HDR images enabled via `SUPPORT_FILEFORMAT_HDR` in the Raylib configuration.

Bigfoot71:
Images embedded in GLB models are normally in JPG format _(that's the case for `DamagedHelmet.glb`)_

Like grunthon said, are you using raylib via the submodule, or your own precompiled version?
It's possible that JPG format support is disabled on your side.

Bigfoot71:
> WARNING: IMAGE: Failed to load image data
> INFO: R3D: Loaded 8 textures successfully

However, I'm bothered by this stupid log, I'll fix that.

sacplex:
I’m using the master branch with the vendored Raylib and Assimp. I made sure SUPPORT_FILEFORMAT_HDR and SUPPORT_FILEFORMAT_JPG are enabled (they were commented out before) in build/src/config.h and did a full clean build with cmake --build . --target clean.

I’m still getting the same warnings and the textures in the PBR example are not loading correctly. I’ve built with VS 2026 in both Debug and Release configurations.

[...]

PR #166, opened 2026-01-22
Set build output directories

grunthon:
Adds consistent build output directories for the library and submodules. This was mentioned in #156, with the issue being that the shared libraries from raylib and assimp would not be placed in the same output directory as the main project. This meant having to manually find and place the shared library files into the appropriate locations. Having the output directories explicitly set also helps with consistency instead of relying on the platform or compiler using default outputs.

This also enables jpg support in raylib since it seems it may be needed.

If another way of handling this would be preferred, let me know.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,16 @@ set(R3D_CONFIG_DEFAULTS
     R3D_ENABLE_TRACELOG             1
 )
 
+# Project output directories
+
+set(R3D_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
+set(R3D_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
+set(R3D_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
+
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${R3D_RUNTIME_OUTPUT_DIRECTORY})
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${R3D_LIBRARY_OUTPUT_DIRECTORY})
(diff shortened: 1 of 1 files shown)


Bigfoot71:
It's perfect, it looks clean to me, thank you!

> This also enables jpg support in raylib since it seems it may be needed.

Yeah, I haven't had time to look into it further, this thing seems strange to me

grunthon:
> Yeah, I haven't had time to look into it further, this thing seems strange to me

I think I see what's going on. In raylib's config.h there is `#define SUPPORT_CLIPBOARD_IMAGE    1` by default, and having that set is enabling jpg support. So jpg is enabled by default even though `SUPPORT_FILEFORMAT_JPG` is commented out.

```
// Support for clipboard image loading
// NOTE: Only working on SDL3, GLFW (Windows) and RGFW (Windows)
#define SUPPORT_CLIPBOARD_IMAGE    1

// NOTE: Clipboard image loading requires support for some image file formats
// TODO: Those defines should probably be removed from here, I prefer to let the user manage them
#if defined(SUPPORT_CLIPBOARD_IMAGE)
    #ifndef SUPPORT_MODULE_RTEXTURES
        #define SUPPORT_MODULE_RTEXTURES 1
    #endif
    #ifndef STBI_REQUIRED
        #define STBI_REQUIRED
    #endif
    #ifndef SUPPORT_FILEFORMAT_BMP // For clipboard image on Windows
        #define SUPPORT_FILEFORMAT_BMP 1
    #endif
    #ifndef SUPPORT_FILEFORMAT_PNG // Wayland uses png for prints, at least it was on 22 LTS ubuntu
        #define SUPPORT_FILEFORMAT_PNG 1
    #endif
    #ifndef SUPPORT_FILEFORMAT_JPG
        #define SUPPORT_FILEFORMAT_JPG 1
    #endif
#endif

```

Bigfoot71:
> I think I see what's going on. In raylib's config.h there is `#define SUPPORT_CLIPBOARD_IMAGE 1` by default, and having that set is enabling jpg support. So jpg is enabled by default even though `SUPPORT_FILEFORMAT_JPG` is commented out.

Aaah, well that explains everything, thank you for that!

PR #169, opened 2026-01-24
Kinematics

Bigfoot71:
Add a module called "kinematics", which provides additional functions on top of raylib for collision and penetration checks, sliding, sweeps, depenetration, ground tests (box/mesh), as well as raycasting functions against R3D mesh data and models, plus a few helper utilities.

The goal is not to provide a full rigid body physics system or anything very advanced, it's mainly intended for prototyping or simple games.

At the moment, the slide/sweep functions only work for capsules, spheres, and boxes against other boxes. I'll look into whether slide/sweep against convex meshes can be implemented in a stable way without too much complexity, but it will remain a fairly simple set of functions. That's why the term **"kinematics"** is used rather than **"physics"**, to avoid misleading users about the capabilities offered.

If anyone comes across this, I'm open to any suggestions!

All of this comes with a new example of about 150 loc:

[Enregistrement d'écran_20260124_211150.webm](https://github.com/user-attachments/assets/fa27df43-dee6-4ac8-9be3-84c5a318f53f)


Bigfoot71:
I've implemented sliding for spheres and capsules against meshes!

I've also simplified things quite a bit by removing box sliding, which honestly wasn't very important and would have made everything much more complicated if it had needed to be as complete as for spheres and capsules.

I also added a function to generate a "slope mesh" which is basically a cut cube where you specify the slope normal / cutting plane from the origin:
```c
R3D_Mesh R3D_GenMeshSlope(float width, float height, float length, Vector3 slopeNormal)
```

Here are two examples:

[Enregistrement d'écran_20260125_234851.webm](https://github.com/user-attachments/assets/57c51ade-f5d2-4d37-914f-8d659fd92fd5)

[Enregistrement d'écran_20260125_235413.webm](https://github.com/user-attachments/assets/8bff46c9-0f90-4b7b-8685-4e2cc4e560c7)

PR #171, opened 2026-01-28
Custom shaders

Bigfoot71:
This PR opens a draft for the current WIP implementation of custom shaders.

Original discussion: https://github.com/Bigfoot71/r3d/discussions/28

The general direction is to split things into:
- **Surface shaders** -> used for materials and decals
- **Screen shaders** -> for screen-space effects (not implemented yet)

For now, only surface shaders are implemented, and the design is still very much open to discussion.

## Current state

What's mostly done so far:

- Shader parsing and reflection
- Optional `vertex` / `fragment` stages (at least one required)
- Support for:
  - Up to 16 uniforms (default, configurable via macros) (bool, int, float, ivec2/3/4, vec2/3/4, mat2/3/4)
  - Up to 4 samplers (default, configurable, regardless of sampler type) (samplers 1D/2D/3D/Cube)
  - Basic varyings

The API exposes an opaque `R3D_SurfaceShader` (name still up for debate) that internally owns all the generated stage shaders.

For now, I preferred this name to `R3D_MaterialShader` because it can also be used for decals.

## Uniform handling

User defined uniforms are written as plain GLSL uniforms, for example:
```glsl
uniform float value;
uniform vec4 color;
```

At load time:
- Analyze uniforms and build a `name -> offset/size` mapping
- Rewrite them into a single `std140` uniform block internally
- Treat all value uniforms as "UBO backed"

On the API side, `R3D_SetSurfaceMaterialUniform()` just writes into a byte buffer in RAM and marks it dirty.
On the next shader use, the buffer is uploaded to the UBO.
This allows sharing the same uniform storage across shaders and avoids redundant uploads.

Samplers stay as regular uniforms.

## Generated code model

User code is split into stages using macros, and rewritten internally into something like:

- A shared uniform block and functions, constants, etc.
- `#ifdef STAGE_VERT` / `#ifdef STAGE_FRAG` sections
- Stage-specific inputs/outputs

At least one `vertex()` or `fragment()` function must be specified, but having both is not mandatory.

This part is mostly functional already.

## Compilation model

Right now, a shader is pre-compiled when loaded, so there is always at least one real compilation happening as validation.

Originally I mentioned lazy compilation, in practice this currently means:
- The main (geometry) shader is compiled first
- Other variants may be compiled later if/when needed

I haven't implemented any usage hint system yet to control which variant should be pre-compiled first.

This would be a good idea to ensure that we compile a shader that will actually be used,
but this is something I'm still unsure about and open to feedback on.


Bigfoot71:
## Open questions / design discussions

Here are a few design points where I'd like input or feedback:

### 1. Naming

Currently these are called **surface shaders** (instead of "material shaders") since decals can also use them.
Open to suggestions if anyone has a better naming idea.

### 2. Color uniforms and sRGB

Should we provide an explicit mechanism, similar to Godot's `source_color` in GDShaders, to indicate that a uniform represents color data and should be converted from sRGB to linear?

This also brings up textures:
- Should we offer an alternative texture loading path (in addition to raylib's) that allows loading textures explicitly in sRGB?
- Or should we leave all of this entirely up to the user?

### 3. Vertex vs fragment data model

Vertices are in local space before transformation, but fragment data is already sampled.
This means you can't really modify texcoords in the fragment stage unless you re-sample manually.

Options to consider:
- Encourage users to handle all sampling themselves
- Or provide more flexibility / helper mechanisms by default

### 4. Shader interface style

Should we stick with Godot-style globals (`POSITION`, `COLOR`, all caps),
or move toward explicit `input` / `output` structs to clearly separate what's read-only versus writable?

### 5. Usage hints

Should we allow specifying the intended usage of a shader so that a "primary" variant can be precompiled early and used as a validation step?

Bigfoot71:
Also, here's the current shader example, just a random program trying to cover most of the small implemented details:
```glsl
uniform sampler2D u_texture;
uniform float u_time;

varying float v_time;

void vertex() {
    v_time = (0.5 * sin(u_time) + 0.5);
    POSITION *= 1.0 + v_time;
}

void fragment() {
    vec2 uv = vTexCoord;
    uv.y += v_time * 0.5;

    const vec3 base_color = vec3(0.5, 0.1, 0.0);
    const vec3 active_color = vec3(1.0, 0.2, 0.0);
    vec3 color = mix(base_color, active_color, v_time);

    ALBEDO = color * texture(u_texture, uv).rgb;
    EMISSION = ALBEDO * v_time * 0.5;

    ROUGHNESS = mix(1.0, 0.25, v_time);
    METALNESS = mix(0.5, 1.0, v_time);
}
```

[Enregistrement d'écran_20260128_121107.webm](https://github.com/user-attachments/assets/3a1ae6ee-88da-4846-a94f-6f7611b88f17)

Bigfoot71:
Testing decals with an ultra minimal test:
```c
const char* code =
    "void fragment() {"
        "ALBEDO *= vec3(fract(gl_FragCoord.x*.1), 0.0, fract(gl_FragCoord.y*.1));"
    "}";

decal.shader = R3D_LoadSurfaceShaderFromMemory(code);
 ```
 
[Enregistrement d'écran_20260128_151415.webm](https://github.com/user-attachments/assets/50370da4-d56c-4052-b2b2-e4b95b0f23c5)

**Comment by grunthon** [comment]
Great! I'll take a closer look at this soon.

Surface shader seems like a reasonable name to me. At least nothing better comes to mind that avoids confusion with materials.

I do like the idea of being explicit with `input` / `output` but I'm not sure it's really needed. If there is any confusion about how the globals should be used that would seem to be an issue beyond just specifying input or output.

Also, just to clarify, when I brought up lazy compilation in the discussion I was referring more to avoiding the possibility of compilation occurring during rendering. That does not seem to currently be the case if I'm reading it correctly, but is related to the idea of usage hints.

Just quick thoughts off the top of my head and the other points I would need to think about more.

**Comment by grunthon** [comment]
In regards to uniform color space conversion and sampling/texture coordinates, in my opinion these should be left to the user to handle. My main reasoning being that trying to deal with this type of stuff in the library is getting towards creating a new/variant shading language. Keeping everything as close to base GLSL as possible would seem to be best for the user and internally for the library.

I could be overthinking it though. I'm thinking more in overall approach than perhaps what might end up being best in practice.

**Comment by Bigfoot71** [comment]
Thank you for your feedback!

> If there is any confusion about how the globals should be used that would seem to be an issue beyond just specifying input or output.

True, I might be overthinking it. It's not a big deal. The idea comes from Godot, which provides editor templates with built-in documentation, so everything is quickly understandable. That's not the case here, but it's fine I think.

> when I brought up lazy compilation in the discussion I was referring more to avoiding the possibility of compilation occurring during rendering. That does not seem to currently be the case

Actually, it does happen, even for all built-in shaders.

When a surface shader is loaded, I first precompile the "geometry" shader as a test. If it fails, we log the error and return NULL. If it succeeds, we assume the shader will work for all other passes, then the others are compiled "on need".

The idea is "you only pay for what you use", no one is going to make a AA game with R3D that dynamically loads 200 variants of the same shader, so performance impact in practice is minimal.

The hint feature would allow the first compilation test to target a shader that will actually be used, instead of defaulting to "geometry". Otherwise, we pay the compilation cost even if it's only used for decals.

**Comment by Bigfoot71** [comment]
> In regards to uniform color space conversion and sampling/texture coordinates, in my opinion these should be left to the user to handle.

For colors, adding a keyword like `source_color` in the uniform declaration and an internal `isColor` flag in R3D isn't expensive or hard to manage, but the key is what's most user-friendly and feel most natural.

The user could do the conversion themselves, but since it's automatically supported for materials, it seems logical to offer it in shaders as well. But I would also be in favor of keeping it as close as possible to GLSL, it's just a matter of consistency, I'm not really sure.

For sRGB textures, I may have overthought this. We can revisit adding a specific sRGB texture loading function later; it's not critical for this PR.

---

Regarding sampling, I mean, global fragment shader values like `ALBEDO` already contain the material extracted value, so they don't need to be resampled and can be modified directly.

If we want to modify UVs, e.g. for a wave effect, resampling would be necessary. Modern drivers likely optimize the first sample through dead code elimination, but we have to trust the driver (heh).

Alternatively, we could leave globals like `ALBEDO` at zero and let the user handle everything, but that would require exposing all textures and material values publicly, and would mean a lot of repetitive code per user shader.

Another option could be a keyword like `do_not_sample` to tell R3D not to sample, or a function `SampleMaterial(uv)` callable in the fragment shader to fill globals automatically as R3D would by default.

Or we can continue with the current approach. I'm open to whichever works best.

**Comment by grunthon** [comment]
> The idea is "you only pay for what you use", no one is going to make a AA game with R3D that dynamically loads 200 variants of the same shader, so performance impact in practice is minimal.

I might say by the same logic there's no reason to not have the shaders compiled up front. As opposed to the built in shaders, new models or effects can certainly consistently come into the rendered view. Personally I'm not going to be okay with stutters from shaders compiling during rendering.

If the hinting system avoids this that would be good but I do wonder if this level of optimization is needed when, as you say, the library isn't really aimed at massive projects where avoiding compiling possibly unneeded shaders would have a large impact.

**Comment by Bigfoot71** [comment]
I understand the concern about avoiding stutters during rendering, that's a valid goal.

However, precompiling all variants means paying an upfront cost for variants that will never be used, and this cost is guaranteed rather than conditional.

There is no perfect solution, but lazy compilation remains the better default, as it scales with actual usage rather than hypothetical possibilities.

That said, we could provide a function to force full or partial compilation when needed, which would be similar to what three.js offers.

**Comment by grunthon** [comment]
> There is no perfect solution, but lazy compilation remains the better default, as it scales with actual usage rather than hypothetical possibilities.

I understand and won't labor the point, it's just something that's a real sticking point for me.

I will mention that one thought I had is possibly having a system to define flags at the top of the shader file to control how it is compiled. It could be nice to have this functionality actually set inside the shader itself.

Most of the other points you bring up I suspect I'll need to actually work with the custom shaders to have any potentially useful thoughts on. Even as it is it's very well implemented.

**Comment by Bigfoot71** [comment]
> I will mention that one thought I had is possibly having a system to define flags at the top of the shader file to control how it is compiled. It could be nice to have this functionality actually set inside the shader itself.

Oooh, yes, that would be a good idea, I'll look into that.

For now, we'll stick with the Godot style, with global variables in all caps, so I'll complete them.
I hadn't wanted to finish it until it was confirmed.

I'll check the rest, thanks for your feedback!

**Comment by grunthon** [comment]
Something I considered waiting until the PR was complete to bring up but I'll go ahead and mention is the use of `typeof` included here. Since the build is configured for C11 you're relying on GCC extensions and `__typeof__` for MSVC which does seem to work but is kind of dancing around the standards versus actually using C23.

This causes (a large amount of) reported errors in Visual Studio where `TYPEOF` is used, despite compiling without errors. It makes working with the code base pretty rough. MSVC has had support for `typeof` since sometime in 2023 and using `clatest` and `typeof` does solve the issue.

You could see this as just yet another quirk of Visual Studio/MSVC but since it originates from using non-standard compiler extensions I wanted to see if there were any thoughts on it beyond any specific IDE or compiler. If you're alright it, that's certainly up to you. Just for an open source project it does impact the ability to work with it.

**Comment by Bigfoot71** [comment]
Alright, I've added the hints in the form of `#pragma` directives directly inside the shaders. Here's an example:
```glsl
#pragma usage opaque shadow
```

This will automatically compile all shaders for the specified usages at load time.
The possible usages are: `opaque`, `prepass`, `transparent`, `shadow`, `decal`, `probe`
Obviously, this remains optional; otherwise, just the "geometry" shader will be compiled to serve as a verification.

---

I've also fixed the "when to sample" issue.

Output values such as `ALBEDO` are always pre-fetched by default, but you can add `#define R3D_NO_AUTO_FETCH`, which disables automatic fetching.

You can then use `FetchMaterial(vec2 texCoord)`, which automatically fills in all values the same way r3d would, but allows you to specify custom UVs.

To go further, I also added these functions:
- `vec4 SampleAlbedo(vec2 texCoord)`
- `vec3 SampleEmission(vec2 texCoord)`
- `vec3 SampleNormal(vec2 texCoord)`
- `vec3 SampleOrm(vec2 texCoord)`

I chose this approach because directly exposing textures and material values made things significantly more complicated.

Now, for example, the `SampleNormal()` function returns only zeros when running in a depth pass.
This is an implementation detail, and unless you're doing something really unusual,
like changing the alpha according to the normal, it should never cause any issues.

Bigfoot71:
> I'll go ahead and mention is the use of `typeof` included here.

Aaaah, I was worried about MSVC, for now the CI wasn't complaining loudly… I thought that was good enough for now.
_But seriously, GCC support this for more than 20 years :')_

Well, if I understand correctly, the real pain point is intellisense, but on the other hand it would be dishonest to claim that r3d is C11 with this anyway.

In practice we can do without `typeof`, it's just that some macros are going to get pretty ugly... :/
Or we could roll a custom `typeof` for those macros using `_Generic`.

I'll look into that, thanks for pointing it out anyway.

grunthon:
> Well, if I understand correctly, the real pain point is intellisense, but on the other hand it would be dishonest to claim that r3d is C11 with this anyway.

Yeah it's intellisense. It also has problems with the atomic types in the texture importer but that's at least pretty minor. As far as I know both of these are known and "being investigated" at the usual Microsoft speed.

> Or we could roll a custom typeof for those macros using _Generic.

I'd thought of that as well but didn't want to get too off track here. I mainly wanted to raise the issue of C version and relying on extensions.

Bigfoot71:
Okay, I removed `typeof` and did some testing, and we have something acceptable for macros (without `_Generic` or anything), you just shouldn't look at the expanded code x)
And yeah, it's more sensitive with threads, we'll have to work with it I guess...

duke4e:
Screenshot 2026-01-29 at 23 25 57

joining the shader party, looks great so far!

Bigfoot71:
Screen shaders have been added!

They are similar to surface shaders, except that there is only one entry point `void fragment()`, and you only get one output `vec3 COLOR`.

The principle is that you can create a "chain" of screen effects using `R3D_SetScreenShaderChain(shaders, count)`, which will be executed at the end of the rendering.

Within these screen shaders, you can fetch/sample scene colors, depth, position, and normal (in viewspace).

And yes, it was already possible to do this by rendering in a render texture and obtaining the internal buffers if necessary, but it was clearly more of a hassle to do, and required another complete framebuffer, here at least it is directly integrated and very simple to do, although it is limited to small effects.

Here's a quick example:
```glsl
uniform float u_time;

void fragment() {
    vec2 uv = TEXCOORD * 2.0 - 1.0;
    uv.x += sin(uv.y * 10.0 + u_time * 3.0) * 0.02;
    uv.y += sin(uv.x * 10.0 + u_time * 2.0) * 0.02;
    COLOR = SampleColor(uv * 0.5 + 0.5);
}
```

[Enregistrement d'écran_20260130_174558.webm](https://github.com/user-attachments/assets/36231b72-8e17-44b9-b5b2-5c239fc1e052)

Bigfoot71:
Alright, I think we're good to go for the review, everything that was on the roadmap is done.

There could definitely be more shaders possible, like for skyboxes or even light shaders, I'll think about that later.

The problem with light shaders is that they could only be available for deferred lighting, unless we get crazy.
As for skyboxes, that's interesting, but I need to think about, the real needs will naturally emerge over time.

Also, the "screen shaders" could go even further, we could imagine a multi-pass system with dynamic buffer management and all that, but the goal isn't to make a super modern, ultra advanced engine.

So overall, the current setup probably covers 90% of the needs and the reality of the projects that will be made with r3d.

grunthon:
I wasn't expecting the screen shaders already. It's great! No problems with brief testing here but I'll try harder to break it soon ;)

A couple of really minor things in the shader example are the torus mesh is `sphere` and the screen shader is not being unloaded.

Bigfoot71:
> No problems with brief testing here but I'll try harder to break it soon ;)

I'll still also run a few tests before merging, but feel free, go all out!
Also, I'll see where to add a bit of doc about the shaders as well.
And yes, thanks for the example, it's fixed!

Bigfoot71:
Okay, I've added some doc for the shaders. I tried to write it so it speaks both to beginners and to people who already know what's going on.

- https://github.com/Bigfoot71/r3d/blob/custom-shaders/docs/shaders/surface_shader.md
- https://github.com/Bigfoot71/r3d/blob/custom-shaders/docs/shaders/screen_shader.md

For screen shaders, I made sure to include a few practical examples, like outlines, vignette, that kind of stuff.

I'll finish a few more tests and do the review tomorrow, and it should be good to go.
If anyone disagrees, speak now or forever hold your peace! :)

Bigfoot71:
Alright, it seems generally fine to me. I was able to fix a few small issues, I think it's good.
Additions can be made gradually as needed. I'll merge!

grunthon:
A great addition to the library!

PR #189, opened 2026-02-08
Fix recursive call loading model from memory

grunthon:
`R3D_LoadModelFromMemory` was calling itself instead of `R3D_LoadModelFromMemoryEx`. I am assuming a value of 0 is correct for the flags when using `R3D_LoadModelFromMemory`.


Bigfoot71:
Oh damn, thanks for that!

PR #195, opened 2026-02-11
SSE checks for MSVC

grunthon:
MSVC does not define any SSE preprocessor macros. This adds checks using the available predefined macros, which as far as I'm aware are only able to determine up to SSE2. Anything beyond that (SSE3+) would need to be manually defined or require a more involved method such as an extra script to test and define them.

Note that the `_M_IX86_FP` checks are only for x86 processors and might be considered unneeded.

`__AVX__` and `__AVX2__` are available and do work correctly so there is no issue with those.

### DIFF
--- a/src/common/r3d_simd.h
+++ b/src/common/r3d_simd.h
@@ -49,13 +49,13 @@
 #endif
 
 #if defined(__SSE2__) || _M_IX86_FP == 2 || (defined(_M_AMD64) || defined(_M_X64)) 
-#define R3D_HAS_SSE2
-#include <emmintrin.h>
+    #define R3D_HAS_SSE2
+    #include <emmintrin.h>
 #endif
 
 #if defined(__SSE__) || _M_IX86_FP == 1
-#define R3D_HAS_SSE
-#include <xmmintrin.h>
+   #define R3D_HAS_SSE
+   #include <xmmintrin.h>
 #endif
 
 #if defined(__ARM_NEON) || defined(__aarch64__)
--- a/src/common/r3d_simd.h
+++ b/src/common/r3d_simd.h
@@ -48,14 +48,14 @@
     #include <pmmintrin.h>
 #endif
 
-#if defined(__SSE2__)
-    #define R3D_HAS_SSE2
-    #include <emmintrin.h>
+#if defined(__SSE2__) || _M_IX86_FP == 2 || (defined(_M_AMD64) || defined(_M_X64)) 
+#define R3D_HAS_SSE2
+#include <emmintrin.h>
 #endif
 
-#if defined(__SSE__)
-    #define R3D_HAS_SSE
-    #include <xmmintrin.h>
+#if defined(__SSE__) || _M_IX86_FP == 1
+#define R3D_HAS_SSE
+#include <xmmintrin.h>
 #endif
 
 #if defined(__ARM_NEON) || defined(__aarch64__)
--- a/src/common/r3d_simd.h
+++ b/src/common/r3d_simd.h
@@ -54,8 +54,8 @@
 #endif
 
 #if defined(__SSE__) || _M_IX86_FP == 1
-   #define R3D_HAS_SSE
-   #include <xmmintrin.h>
+    #define R3D_HAS_SSE
+    #include <xmmintrin.h>
 #endif
 
 #if defined(__ARM_NEON) || defined(__aarch64__)


grunthon:
There is a possible discussion around if it would be better to determine and use the correct SIMD functions at runtime rather than compile time. Runtime would seem to be better for the actual end user of the application. It would also avoid relying on inconsistent macro definitions plus doing CPU checks at initialization could be used for this as well as core/thread count for loading.

Bigfoot71:
Aïe, I hadn't thought of that. I literally grabbed that code from one of my projects, and I didn't think to test it for Windows, which is really stupid. Nothing has changed since...

And yes, I completely agree that it should be resolved dynamically.
I'll look into it! And thanks for this fix!

grunthon:
I put this together using `cpuid` to get the needed info. It needs testing but I might mess around with it more to see if it's viable.

```
#if defined(_MSC_VER)
    #include <intrin.h>
#elif defined(__GNUC__) || defined(__clang__)
    #include <cpuid.h>
#endif

void CPUInfo(int info[4], int function) {
#if defined(_MSC_VER)
    __cpuid(info, function);
#elif defined(__GNUC__) || defined(__clang__)
    __get_cpuid(function, &info[0], &info[1], &info[2], &info[3]);
#else
    __asm__ volatile (
        "cpuid"
        : "=a" (info[0]), "=b" (info[1]), "=c" (info[2]), "=d" (info[3]) // outputs
        : "a" (function) // input
        );
#endif
}

void CheckCPU() {
    int info[4];
    CPUInfo(info, 1);

    if (info[3] & (1 << 25)) printf("SSE supported\n");
    if (info[3] & (1 << 26)) printf("SSE2 supported\n");
    if (info[2] & (1 << 0))  printf("SSE3 supported\n");
    if (info[2] & (1 << 9))  printf("SSSE3 supported\n");
    if (info[2] & (1 << 19)) printf("SSE4.1 supported\n");
    if (info[2] & (1 << 20)) printf("SSE4.2 supported\n");
    if (info[2] & (1 << 28)) printf("AVX supported\n");
    if (info[1] & (1 << 5)) printf("AVX2 supported\n");

    CPUInfo(info, 0x0A);
    int cores = (info[0] & 0xFF) + 1;

    printf("Logical processors: %d\n", cores);
}
```

Bigfoot71:
Oh cool, thanks! I had just started, for now just with `__get_cpuid`, I'll test your code, I was reviewing matrix functions, it going well for now!

grunthon:
I'm not sure about the inline assembly for other compilers, or if it's even really needed.

Also this is (should be?) getting the actual physical core count which may or may not be better to use for a thread amount versus the virtual core count that is being used now.

Bigfoot71:
> I'm not sure about the inline assembly for other compilers, or if it's even really needed.

Yeah, I don't really see either how that would be useful. TinyCC maybe, but hey

> Also this is (should be?) getting the actual physical core count which may or may not be better to use for a thread amount versus the virtual core count that is being used now.

And no, especially since it complicates things, for ARM (including Mac) we don't need all this for SIMD, so it's better to go through the system no matter what.

---

Finally, I got disptach working for MSVC and GCC/Clang, but while batch matrix multiplication with dispatch is still faster than scalar O3 in isolated tests, in r3d conditions it's about the same, or slightly worse, probably due to inlining.

I have ways in mind to make it useful, but they require too many changes for very specific cases. I originally added it for a personal need when the animation system handled everything in one place and was simpler, but now making dispatch worthwhile would be hard to justify.

So the most reasonable approach is to just clean it up and go back to something simpler.
If the need arises again, I'll revisit it differently.

grunthon:
> And no, especially since it complicates things, for ARM (including Mac) we don't need all this for SIMD, so it's better to go through the system no matter what.

That makes sense. I would still be curious to do some profiling comparing using physical core count versus virtual core count for the number of threads.

As far as SIMD, I don't know if you made more progress but my initial thought was to do the extension check during `R3D_Init` and set function pointers to appropriate functions. This would mean breaking the SIMD using functions into separate extension specific ones. It's possibly a bit messy but avoids having to do a lot of conditional checks all the time. Just let me know if you've already figured something out or have a better way to handle it.

> Finally, I got disptach working for MSVC and GCC/Clang, but while batch matrix multiplication with dispatch is still faster than scalar O3 in isolated tests, in r3d conditions it's about the same, or slightly worse, probably due to inlining.

Interesting. I was really only thinking in regards to the texture importer. I assume you mean for `r3d_matrix_multiply_batch`. Are you running it on a single separate thread or actually running a thread pool internally in `r3d_matrix_multiply_batch`? I can see it getting pretty complicated and the benefit (or lack of) being very situationally dependent.

Bigfoot71:
> I would still be curious to do some profiling comparing using physical core count versus virtual core count for the number of threads.

Yeah, profiling that could be interesting. For production though, we need to follow what the OS schedules, using physical cores wouldn't be portable or reliable as a default.

> As far as SIMD, I don't know if you made more progress but my initial thought was to do the extension check during `R3D_Init` and set function pointers to appropriate functions.

Yeah, I wouldn't go with function pointers here. I imagine they only make sense if you have a large set of heavy functions.

On MSVC, I see that you can use `__declspec(cpu_dispatch / cpu_specific)` for dispatch, but there's no equivalent on GCC/Clang, there it's always manual, either via separate files or function attributes.

I did some tests with `r3d_matrix_multiply_batch`, but results were too variable and the complexity wasn't worth it. To avoid dealing with all that, I removed SIMD from these functions on master, as I said it was just legacy from an old project and just a headache in the end, I'll look at that later if really needed again.

> Interesting. I was really only thinking in regards to the texture importer. I assume you mean for `r3d_matrix_multiply_batch`. Are you running it on a single separate thread or actually running a thread pool internally in `r3d_matrix_multiply_batch`?

However, the texture importer didn't use any hand written SIMD, and `r3d_matrix_multiply_batch` wasn't parallelized.  

I did consider using SIMD for the ORM composition in the importer, but it would have required rewriting a lot of stuff that raylib already provides, so I never did it, though I might have mentioned it at some point.

PR #197, opened 2026-02-14
Unified forward and deferred shadow map sampling functions

grunthon:
Replaces the mostly duplicated shadow sampling functions in forward and deferred rendering with a new shared included block. The main purpose being to eliminate duplicate code to make maintenance and customizing shadow sampling easier.

Any feedback welcomed.

### DIFF
--- a/shaders/deferred/lighting.frag
+++ b/shaders/deferred/lighting.frag
@@ -14,9 +14,6 @@
 
 /* === Includes === */
 
-#include "../include/blocks/light.glsl"
-#include "../include/blocks/shadow.glsl"
-#include "../include/blocks/view.glsl"
 #include "../include/math.glsl"
 #include "../include/pbr.glsl"
 
@@ -34,6 +31,12 @@ uniform sampler2D uOrmTex;
 uniform sampler2DArrayShadow uShadowDirTex;
 uniform sampler2DArrayShadow uShadowSpotTex;
 uniform samplerCubeArrayShadow uShadowOmniTex;
+
+/* === Blocks === */
+
+#include "../include/blocks/light.glsl"
+#include "../include/blocks/shadow.glsl"
+#include "../include/blocks/view.glsl"
 
 /* === Fragments === */
 
@@ -100,10 +103,7 @@ void main()
 
     /* --- Calculating a random rotation matrix for shadow debanding --- */
 
-    float r = M_TAU * M_HashIGN(gl_FragCoord.xy);
-    float sr = sin(r), cr = cos(r);
-
-    mat2 diskRot = mat2(vec2(cr, -sr), vec2(sr, cr));
+    mat2 diskRot = S_DebandingRotationMatrix();
 
     /* Apply shadow factor if the light casts shadows */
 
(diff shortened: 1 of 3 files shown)


Bigfoot71:
Yeah, this bothers me too. My main concern is passing the `Light` struct as a parameter, I'm not sure what the real impact is. I seem to recall a case where passing a struct as a parameter versus using a global was measurably slower.

It would be great if we could inspect the driver generated code to see whether the call gets inlined _(even if that's only meaningful per driver)_, or even compile it ourselves, but yeah...

That said, it probably needs some cleanup anyway, and the light was redefined locally in the forward shader, but anyway... If you have an opinion on this, I'm all ears.

Also, it looks like the same rotation matrix used for jittering is recomputed per light in the forward shader.

grunthon:
> My main concern is passing the Light struct as a parameter

There's really no functional reason either the light or sampler have to be passed since it's just being directly included in the the shaders. I was mainly trying to keep the functions more abstracted from the main shaders.

> Also, it looks like the same rotation matrix used for jittering is recomputed per light in the forward shader.

Ah yes you're right.

> It would be great if we could inspect the driver generated code

Yeah I thought the same and might look at doing this but as you say it doesn't mean a huge amount outside the specific driver. I suspect any difference would be extremely small and wouldn't outweigh the benefit of the code being better to work on, but I could be wrong.

I'll make some changes and you can see what you think.

Bigfoot71:
> There's really no functional reason either the light or sampler have to be passed since it's just being directly included in the the shaders. I was mainly trying to keep the functions more abstracted from the main shaders.

I may be misunderstanding, but my concern is that the `Light` struct is now passed as a function parameter.
It might be copied on every call, unless it's inlined or something, but generally it's best to avoid it.

Maybe we could look at all this differently, by splitting the logic in another way.

grunthon:
My first thought was to not pass the `Light` and choose the appropriate one based on if the function was being used in the forward or deferred shader. It does still need the index passed for the forward shader which is obviously useless for deferred.

```
float S_SampleShadowDir(vec4 projPos, float cNdotL, int i)
{
    Light light;
#ifdef NUM_FORWARD_LIGHTS
    light = uLight[i];
#else
    light = uLight;
#endif
...
```

This is just off the top of my head.

I suppose the function parameters could be different per shader as well, but that's getting a little odd.

```
#ifdef NUM_FORWARD_LIGHTS
float S_SampleShadowDir(vec4 projPos, float cNdotL, int i)
    Light light = uLight[i];
#else
float S_SampleShadowDir(vec4 projPos, float cNdotL)
    Light light = uLight;
#endif 
```

Bigfoot71:
Yeah, I think we could be even more aggressive:
```glsl
#ifdef NUM_FORWARD_LIGHTS
float S_SampleShadowDir(vec4 projPos, float cNdotL, int i)
{
    #define light uLight[i]
#else
float S_SampleShadowDir(vec4 projPos, float cNdotL)
{
    #define light uLight
#endif

    /*...*/
    
    #undef light
}
```

----

Aaah, you made almost the same edit while I was writing it x)
Otherwise, the idea of using `#define` for indexing is probably better than a local copy.
Honestly, just make the adjustment that seems the healthiest to you, I'll be fine with it!
I'll revisit it if needed later.

grunthon:
I'll mention this here for the moment, but I did run into textures intermittently not loading or rendering. I was able to reproduce it in master as well. I want to test some more and make sure it's not something on my end before looking further or creating an issue.

missing-inmaster
missing-inmaster2

grunthon:
I have updated the parameters along the lines discussed and the rotations for sampling should be correct now.

Bigfoot71:
It was good to pass the `mat2` as a parameter, as I mentioned, otherwise it would be recomputed identically for each light in forward.

I had also forgotten to mention that the functions could be directly in `blocks/light.glsl`, since `blocks/shadow.glsl` suggests that a uniform block is defined there, that may not have been clear enough.

Overall this is better, thanks! I'll make the adjustments.

---

> I'll mention this here for the moment, but I did run into textures intermittently not loading or rendering. I was able to reproduce it in master as well. I want to test some more and make sure it's not something on my end before looking further or creating an issue.

I also ran into this issue on Sponza yesterday. It only happened once for me, but it does confirm the problem. I'll look into it as soon as I'm done with what I'm working on now.

grunthon:
> It was good to pass the mat2 as a parameter, as I mentioned, otherwise it would be recomputed identically for each light in forward.

Ah I was thinking it was probably best to have random rotation per light while missing the obvious fact that it would still be the same since it is based off `gl_FragCoord`. Which of course makes sense for stability.

> I had also forgotten to mention that the functions could be directly in `blocks/light.glsl`

I had misinterpreted the meaning of blocks in the context of the directory name but I see now. I actually did have them in `light.glsl` originally.

PR #200, opened 2026-02-15
SSGI Implementation

Bigfoot71:
This PR adds a SSGI implementation and refactors SSIL along with its environment parameters.

**SSGI Implementation:**
The new SSGI is currently based on simple linear raymarching (optimizations will come over time) rendered at half resolution.

The idea, like with SSR, is straightforward, shoot N rays from each pixel to find diffuse light sources illuminating it. 

The current detailed approach is:
- For each pixel, we generate a ray direction over the hemisphere (tangent space) using an interleaved 4x4 pattern (azimuth + elevation rings) for maximum spatial stability.
- We then apply a tile-quantized rotation derived from a rank-1 lattice (bits of `tileCoord`) plus a fixed step per sample, which "fills" the disk without visible random noise.
- The ray is marched in constant steps up to `uMaxDistance`, projected to UV at each step, and we detect intersections via a ray-depth vs scene-depth test with a thickness `uThickness` (first hit).
- We accumulate the radiance at the hit point (`diffuse + history`) with a distance fade, then average over `uSampleCount`.

**SSIL Refactor:**
SSIL has been reworked with simplified parameters and some bug fixes. I hesitated to keep it since it might seem redundant with SSGI + SSAO, but honestly, after extensive testing, we can mix all three together while maintaining acceptable framerates, and the combination produces convincing results.

The key takeaway, SSIL with a low radius works well with SSAO, while a higher radius pairs beautifully with SSGI to refine details.

---

[Enregistrement d'écran_20260215_211246.webm](https://github.com/user-attachments/assets/3eab96c2-842f-48b9-a9a9-09de9e7f00e9)


Bigfoot71:
Without any possibility of temporal reprojection (a current design limitation of R3D), the idea seems to be to achieve maximum spatial stability, and obviously, with a good distribution free of repetitive patterns.

I haven't pushed the current changes yet because I need to find a reliable way to properly correlate all the parameters, as what I have so far depends on the number of samples.

But currently, here's an example with 8 samples:

[Enregistrement d'écran_20260216_231842.webm](https://github.com/user-attachments/assets/32a20be4-1287-4fe9-8c40-94d1053d698c)

And the result once blurred:

[Enregistrement d'écran_20260216_231758.webm](https://github.com/user-attachments/assets/2eb8b004-1dd5-492a-a900-ed78b7e64bb6)

What you see is the contents of the SSGI buffer, the scene is a flat white floor receiving direct light, with a red emissive sphere above it, and I rotate the camera around it, you can see that it is very stable.

One or two samples work very well too, once blurred and composed with the scene, the stability remains the same. I used eight for this example because it's more visually understandable, and especially because the noise is hard on the eyes after a while x)

And here's what it looks like with 2 samples:

[Enregistrement d'écran_20260216_233544.webm](https://github.com/user-attachments/assets/185d8db9-2d76-48c6-98c5-ab68e0a47285)

Bigfoot71:
Okay, so I managed to get something cool and properly configurable.

Here's a test, with the default configuration: 2 samples per pixel and 5 denoising steps, for an average of ~0.4ms at half resolution for 720p, and ~1.5ms at 1080p _(more optimizations will follow this PR)_.

I wrote a short post here to explain a little about the why/how of the approach:
https://bigfoot71.github.io/2026/02/17/ssgi-without-temporal-reprojection/

---

[Enregistrement d'écran_20260217_163434.webm](https://github.com/user-attachments/assets/10e96e59-3397-4247-bbdf-a7211eda2d0b)

grunthon:
This is looking quite nice and will dig more into it soon.

One nice simple addition would be a basic intensity/energy multiplier value.

Bigfoot71:
> One nice simple addition would be a basic intensity/energy multiplier value.

Yeah, I thought about it, but I didn't like the results I was getting, it felt strange.
That said looking back, my opinion doesn't matter, it should definitely be a freedom.

PR #201, opened 2026-02-16
Thread safe image format hint

grunthon:
Using Raylib's `TextFormat` was causing problems when creating the image format hint to pass to `LoadImageFromMemory` when using multiple threads. `TextFormat` keeps a fixed number of buffers and these were being overwitten when the number of threads using it to create strings exceeded the amount of buffers being used.

Feel free to handle this in a different manner, but this was the cause of textures not loading correctly when using `R3D_LoadModel`.

### DIFF
--- a/src/importer/r3d_importer_texture.c
+++ b/src/importer/r3d_importer_texture.c
@@ -13,6 +13,7 @@
 #include <assimp/material.h>
 #include <assimp/texture.h>
 
+
 #ifdef _WIN32
 #   define NOGDI
 #   define NOUSER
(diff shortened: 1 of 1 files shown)


Bigfoot71:
Ah yes, I should have paid more attention to that!
Just, `aiTex->achFormatHint` is fixed-size, defined by `HINTMAXTEXTURELEN`, so `malloc` can be avoided.
See: https://github.com/assimp/assimp/blob/master/include/assimp/texture.h#L148-L167
Let me know if you'd prefer I handle it.

grunthon:
Ah I had not seen that. Updated.

Bigfoot71:
Thanks!

PR #209, opened 2026-02-22
add new module R3D_AnimationTree

misokobza:
- initial version
- missing API docs/comments


misokobza:
Hi @Bigfoot71, 
thanks for help, I managed to make a PR! :)

All planed features are coded, but of course they need more testing.

What's missing:
- R3D_AnimationTree API description (file include/r3d/r3d_animation_tree.h)
- Better example. I would like to create new animation file, with some mannequin and only actually used animations in the example. Also, I would like to create some simple but explanatory animation tree, maybe with utilized bone mask? Let me pls know what do you expect from the example.

Open questions:
- There are some similar (or fully the same) functions in src/modules/r3d_animtree.c and src/r3d_animation_player.c, like compute_model_matrices and functions working with animation channel. I was not sure what to do with it, and didn't want to create new file/module.
- Pls check functions transform_add_v and transform_addx_v in src/modules/r3d_animtree.c. I have originally used add_v version for switch node, which is basically the same logic as animation player, but some animations were not correctly blended. Rotations were not following the shortest path, and for example the legs of character were doing "wheels" sometimes. I have intuitively fixed it with quaternion lerp , and all looks fine now. I'm not sure if animation player should be fixed too, but from what I've tested, using QuaternionAdd doesn't always work correctly (regardless of normalization).

Let me know what you think, I'm open to change/modify anything and everything. :)

Have a nice evening.

Bigfoot71:
This looks great! I took my time reading through everything.

The main issue I see is the placement of code in `modules/r3d_animtree.c`. The `modules/` directory is intended for stateful modules that require initialization/deinitialization, which isn't the case here.

Here's what I'd suggest instead:
- Create `common/r3d_anim.h` and `common/r3d_anim.c` for any functions that can be shared between `r3d_animation_player.c` and `r3d_animation_tree.c`
- For logic tightly coupled to the animation tree, just declare and define those functions as `static` directly in `r3d_animation_tree.c` with simple names

This should also address your open question about duplicated functions and the animation channel helpers.

---

> Rotations were not following the shortest path

I had anticipated this issue but hadn't run into it yet with the animation player, so I had deferred it. Since I was planning to revisit this anyway, unless you've specifically reproduced it with the animation player on a particular model/animation, it would make sense to put those potentially reusable functions `common/r3d_anim.h` and I can then handle the animation player review to reuse them as needed. That said, feel free to tackle it yourself if you're up for it, but it's definitely not required.

---

> Let me pls know what do you expect from the example.

For now I prefer minimal examples, I'm planning to create a dedicated examples repository soon with more polished and better organized, all current examples will be revisited at that point. So there's no need to invest too much time here for now. If you want to review the current one, I'd keep it under ~200 loc, otherwise I'd leave it as is for now.

I think the API is clear enough for anyone who knows what they want to do. The "tutorial" aspect will come with the examples repo and the more complete doc I plan to write, like I've started doing for custom shaders.

misokobza:
Hi @Bigfoot71,
pls check latest commit, if it's what you had imagined. I have updated Animation Player too (but notice the `// TODO:` in the code).

Open question:
- Functions and structs moved from `src/module/r3d_animtree.c` do not follow camel case naming convention for local variables, which was correct for module directory, but not for the main API files, if I checked correctly. Should I rename them all? It will take some time, I don't want to do that automatically, but final code readability will probably change a little (maybe only for me, I'm not sure, I'm not a big fan of camel case :)

Meanwhile, next steps:
- create API header documentation (file `r3d_animation_tree.h`),
- prepare debug test app for animation player, I'm 99.9% sure the rotation issue is there, I just need to find the right animations to lerp,
- update animation tree example.

Bigfoot71:
> * Functions and structs moved from `src/module/r3d_animtree.c` do not follow camel case naming convention for local variables, which was correct for module directory, but not for the main API files, if I checked correctly. Should I rename them all? It will take some time, I don't want to do that automatically, but final code readability will probably change a little (maybe only for me, I'm not sure, I'm not a big fan of camel case :)

No worries about that, I noticed it but didn't point it out, I can clean up the style after the merge, it doesn't bother me.

Only internal functions are in 'snake_case', everything else has remained 'camelCase' so far. If you spotted a snake case variable somewhere, that might just be habit on my side, but it should be fixed for consistency.

Also, only non static functions need to be prefixed with `r3d_`, for static functions we can keep things simple.

_Honestly, from the start I had a questionable mix of snake case internally and camel case publicly for consistency with raylib. At some point I decided that only internal functions would use snake case. This project has been a lot of accumulated struggle, I didn't expect to go this far, but I prefer snake case too ^^_

---

Otherwise, it looks very good, I'll do a full review once you feel it's ready to be merged, but it's looking great so far! :)
And yes, that's exactly what I imagined, no problem!

Regarding the add_v/addx_v in the animation player, yeah, it's clearly shaky. I mentioned in a previous edit to leave it as is for now, but if you find an obvious incorrect case, feel free to correct it, otherwise I could handle it myself later, I already flagged it.

misokobza:
> If you spotted a snake case variable somewhere, that might just be habit on my side, but it should be fixed for consistency.

What a coincidence, when I was originally checking, I've first opened file `r3d_driver.c`, and there is a variable `num_extensions` in `bool r3d_driver_check_ext(const char* name)` function. Now I did more careful check on other files, and looks like it's most probably the only variable with snake case. :D

> Also, only non static functions need to be prefixed with r3d_, for static functions we can keep things simple.

Yup, the code should follow this rule already, if I made no mistake.

> Honestly, from the start I had a questionable mix of snake case internally and camel case publicly for consistency with raylib.

I actually like this idea, I think it was a good approach. But I'll of course follow any given rules, consistency is the most important factor.

> Regarding the add_v/addx_v in the animation player, yeah, it's clearly shaky. I mentioned in a previous edit to leave it as is for now, but if you find an obvious incorrect case, feel free to correct it

Like I wrote, I'll try to prepare a test, I'm sure there are some animations in Human.glb file that show this issue (and addx_v should fix it).

> Otherwise, it looks very good, I'll do a full review once you feel it's ready to be merged, but it's looking great so far! :)

Thank you sir! :) I'll try to finish it this week. I should have some free time, it's spring break here...

misokobza:
Hi @Bigfoot71,
pls check latest commit, this should be it, I have nothing more to add, at least for now... :)

There is a test example file `animation_issue.c`,  demonstrating the AnimationPlayer rotation issue. Just change 'add_v' function to 'addx_v' in `r3d_animation_player.c`, and the issue will be fixed. Pls then remove this test from the PR, or let me know and I'll do it.

Waiting for your feedback, meanwhile I can start with the snake to camel case conversion if you want.

Bigfoot71:
That's perfect! I was able to test it and yeah, it was clearly broken, thanks a lot for the fix!
I removed the test file and took the liberty of uncommenting `addx_v` and removing the todo.

If that works for you, I can do a full review tomorrow. I won't bother you with minor details, and I might clean up some style things, I'll let you know only if I have real doubts, otherwise it's looking good to be merged!

Also don't forget to pull the latest changes if you are planning to make changes

misokobza:
> If that works for you, I can do a full review tomorrow. I won't bother you with minor details, and I might clean up some style things, I'll let you know only if I have real doubts, otherwise it's looking good to be merged!

Awesome, clean whatever is needed, I'll keep an eye on the PR until it's merged. Btw I've removed the test file from make list, it would not compile on my end.

src/r3d_animation_tree.c
```diff
@@ -0,0 +1,1378 @@
...
+{
+    unsigned char in_count  = node->in_cnt;
+    unsigned char active_in = node->params.activeInput;
+
+    for(int i = 0; i < in_count; i++)
+        node->in_weights[i] = 0.0f;
+    node->in_weights[active_in] = 1.0f;
+
+    for(int i = 0; i < in_count; i++)
+        anode_reset(node->in_list[active_in]);
```
Bigfoot71: Shouldn't it be `anode_reset(node->in_list[i]);`, but why two loops?
_(in `anode_reset_switch`)_

I'm not entirely sure, let me know, I've already started the refactoring, I can make the change if needed.

Bigfoot71:
> Btw I've removed the test file from make list, it would not compile on my end.

Thanks about that, I rushed it a bit.

---

I've updated the style to better match with the rest, there shouldn't be any snake case variables left, and I removed the alignments (a bit reluctantly, but they really stood out compared to the rest).

I also made a few adjustments to the evaluation functions and to `atree_update`. It's purely stylistic, the logic is strictly identical (unless I've done something crazy).

The only real issue I've noticed so far is in `anode_reset_switch`, where the same node gets reset in a loop. I'm waiting for your feedback on that.

If you have any concerns about what I've done, feel free to let me know. Otherwise, I'll do one final re-read on my side to be safe, and then I'll merge it!

src/r3d_animation_tree.c
misokobza: This is bug, thx for noticing... fix is commited.

Btw: first loop is for weights, second to reset all connected inputs.

misokobza:
> I've updated the style to better match with the rest, there shouldn't be any snake case variables left, and I removed the alignments (a bit reluctantly, but they really stood out compared to the rest).

Ah yes, the alignments, sorry about that, it really helps me with code readability... and it's a little relaxing process to align them, when I'm thinking about the code. :) I'll follow the style next time.

> I also made a few adjustments to the evaluation functions and to atree_update. It's purely stylistic, the logic is strictly identical (unless I've done something crazy).

Just checked it, I think it should be correct.

> The only real issue I've noticed so far is in anode_reset_switch, where the same node gets reset in a loop. I'm waiting for your feedback on that.

That was a bug, should be fixed now (good eye!).

> If you have any concerns about what I've done, feel free to let me know. Otherwise, I'll do one final re-read on my side to be safe, and then I'll merge it!

No concerns, all looks good!
I'll keep testing the tree in upcoming weeks, going to do some simple IK on model's spine via tree callbacks...

misokobza:
Hi @Bigfoot71, I don't want to do any more changes in code, since final re-read is in process, but I have notice some styling artefacts:
- Function `expand_path` has snake case variable `e_idx`.
- Functions `anode_eval_blend2`, `anode_eval_add2` and` anode_eval_switch` have inconsistent alignment of `.motion` and `.distance` parameters.
- Function `anode_update_stm` has unnecessary space character after `.yes` parameter of `r3d_stmvisit_t`

Bigfoot71:
I found a bug when two non-looping animations are chained in a STM with `R3D_STM_EDGE_ONDONE`. After the first animation ends, the model flashes to T-pose and the tree fails with:

```
WARNING: R3D: Failed to update stm: incorrect time calculation (-0.006171)
ERROR: R3D: Animation tree failed
```

The cause seems to be in the non-looping branch of `anode_update_anim`, `tDelta` is negative when `speed > 0` (since `tCur > boundary`), so this produces a `consumedTime > elapsedTime`, which makes the remaining time negative in the STM:

```c
elapsedTime = elapsedTime * (1.0f - tDelta/tInc);  // consumedTime > elapsedTime
```

I've made this, derived from `consumedTime = (boundary - t0) / speed`:

```c
// Also fixed: tInc != 0 instead of tInc > 0, to handle speed < 0 as well
elapsedTime = (!FloatEquals(tInc, 0.0f) ? elapsedTime * (1.0f + tDelta/tInc) : 0.0f);
node->root.loops = 0;  // non-looping: single boundary crossing, no full cycles
```

I'm getting a bit tired, I'd prefer to have your opinion on this before pushing anything.

Here is a video example:

[Enregistrement d'écran_20260227_224205.webm](https://github.com/user-attachments/assets/cbb0bd5e-dc44-403b-a863-9f5af3d43263)


---

As a side note, in the looping branch there may be a pre-existing issue where `currentTime -= copysignf(duration, speed)` only subtracts one cycle, which would be incorrect if `tInc > 2 * duration`, but that's another thing.

Bigfoot71:
> Hi @Bigfoot71, I don't want to do any more changes in code, since final re-read is in process, but I have notice some styling artefacts:
> 
>     * Function `expand_path` has snake case variable `e_idx`.
> 
>     * Functions `anode_eval_blend2`, `anode_eval_add2` and` anode_eval_switch` have inconsistent alignment of `.motion` and `.distance` parameters.
> 
>     * Function `anode_update_stm` has unnecessary space character after `.yes` parameter of `r3d_stmvisit_t`

Ah, thank you, I'll fix that. I'll get back to it tomorrow evening, hoping we can merge it! :)
Feel free to edit anything in the meantime if you see anything.

misokobza:
> I found a bug when two non-looping animations are chained

Yes I noticed it too, and pushed fix like a minute before your comment. :) Can you pls check if it's fixed? I just found it by reading the code, not by testing. I have swapped variables when calculating `tDelta`, stupid mistake, I'm sorry, now it should be correct for both positive and negative speed. For speed zero, pls see my question at the bottom.

> As a side note, in the looping branch there may be a pre-existing issue where currentTime -= copysignf(duration, speed) only subtracts one cycle

That should be fine, it doesn't matter for the animation's state `currentTime` how many times animation looped. What's important is a would be taken motion of root bone, and that's stored in `node->root.loops`, which is based on elapsed time. Also, this line:
```
node->root.loops = 0;  // non-looping: single boundary crossing, no full cycles
```
should not be necessary, `loops` should be set to zero anytime the animation cross, and elapsed time is shorter than animation duration, so this existing line should handle it all:
```
node->root.loops = (int)(elapsedTime / duration);
```

There is one open question: 
How to behave when animation speed is set to `0.0f`? Currently, animation node reports consumed time equal to elapsed time, but maybe it should report consumed time `0.0f` - this may lead to cycles in state machine, but maybe that's correct behavior if user sets animation speed to zero... lets discuss tomorrow.

Note: after thinking about it, you're right, if speed is zero, consumed time should be zero too. It's the same as if the `play` parameter would be `false`.

Bigfoot71:
> Yes I noticed it too, and pushed fix like a minute before your comment. :) Can you pls check if it's fixed? I just found it by reading the code, not by testing.

Yep, confirmed fixed!

> That should be fine, it doesn't matter for the animation's state `currentTime` how many times animation looped. What's important is a would be taken motion of root bone, and that's stored in `node->root.loops`, which is based on elapsed time.

Ok thanks, that makes sense, that was just a passing thought

> Note: after thinking about it, you're right, if speed is zero, consumed time should be zero too. It's the same as if the `play` parameter would be `false`.

Agreed, that's exactly what we should expect.

---

However, one more issue I just ran into, when chaining two animations with negative speed, the first plays correctly in reverse but the transition to the second never happens. This seems related to the note added in your last commit.

I had a look and it doesn't seem entirely straightforward, it's also a fairly specific case. I'll leave it to your judgement, if there's no obvious fix right now, documenting it clearly somewhere in the code in the meantime seems reasonable.

One last thing, there's no malloc failure checks, but for `atree_stm_create` I think we could also potentially allocate everything in a single block and distribute pointers from there, which would simplify error handling too, but that's polish, so maybe a bit early.

Other than that, this looks good to me, it seems functional for the majority of use cases.
I will wait for your response regarding these last two things, but that's fine with me.

misokobza:
> However, one more issue I just ran into, when chaining two animations with negative speed, the first plays correctly in reverse but the transition to the second never happens.

Could you pls double check whether it's the first animation that is played, and not the second one? What may happen, when animation speed is negative, is that initial active node/state of STM is not reset, and `currentTime` of associated animation is `0.0f`. First update of that node will return `anodeDone` as `true` (baceuse of cross), and STM will immediately jump to next state.

Can you pls try to uncomment that `// TODO` with `anode_reset`, and try whether the issue persist? This wont be the fix, I just want to make sure it is the cause. If so, I have some idea how to approach it in more general way...

Bigfoot71:
> Could you pls double check whether it's the first animation that is played, and not the second one?

Ah yes, sorry, it is indeed the second animation that is playing.

> Can you pls try to uncomment that `// TODO` with `anode_reset`, and try whether the issue persist?

And yes, all the animations play correctly when the reset is uncommented, it starts and ends with the correct ones.

misokobza:
Oka, perfect!

My suggestion is to add new API function ` void R3D_InitializeStmNode(R3D_AnimationTreeNode* node);`, which should be called by user at the end of creating the tree. The function just calls `anode_reset(node);`.

What do you think?

I would also add to the docs, that this function is necessary only when working with negative speeds in STM - it's actually only a problem when initial node has negative speed, other STM nodes are always reset by `stm_next_state`. Or maybe it's too much details for API docs, your call... :)

Bigfoot71:
> My suggestion is to add new API function ` void R3D_InitializeStmNode(R3D_AnimationTreeNode* node);`, which should be called by user at the end of creating the tree. The function just calls `anode_reset(node);`.

As I understand it, the concern is that `anode_reset` could be problematic on incomplete nodes, but would it be possible to guard against that and/or trigger a reset from `R3D_AddRootAnimationNode`, maybe conditionally?

I probably don't have all in mind, just throwing it out there, your solution works fine too if we want something functional, so go for it if that's the best!

As for the docs, feel free to add as much detail as needed, as long as it doesn't directly expose internal concepts or names, otherwise yes, it might get confusing x)

misokobza:
Pls check latest commit. I was thinking about it, and thought it would be maybe better to avoid issues with resetting incomplete nodes all together, by handling it from opposite side.

There is new API function `R3D_CreateAnimationNodeEx`, with one extra boolean parameter to specify whether to set animation playback time automatically (and override `currentTime` value from params). When user is creating animation nodes for State Machine with negative speed, the `Ex` version should be used with `setTime` flag set to `true`.

What do you think?

Bigfoot71:
Oh yes, that's definitely better! _(and confirms that it works)_

But unless there are other uses for it, why not automate the check?
```c
    if (params.state.speed < 0.0f) {
        anode_reset_anim(anim);
    }
```

---

Well, apart from that, everything is really good, thank you very much for your time, I'll do the final reading tonight just for typos or minor things I can fix directly, I think I've tested all the obvious things, thanks again!

misokobza:
> But unless there are other uses for it, why not automate the check?

Hmm, well, because I would like to keep possibility for user to start animation in the middle for example, regardless of animation speed. I don't want to override `currentTime` value from provided `params`, just because of negative speed (and only when negative speed). It should happen only if explicitly allowed by flag.

misokobza:
> Well, apart from that, everything is really good, thank you very much for your time, I'll do the final reading tonight just for typos or minor things I can fix directly, I think I've tested all the obvious things, thanks again!

You are very welcome! It was fun to work with you, and I'm honored to have my code included to this awesome library. I'm going to use R3D in my indie project, so you'll probably hear from me here a lot in future... :)

> One last thing, there's no malloc failure checks, but for atree_stm_create I think we could also potentially allocate everything in a single block and distribute pointers from there, which would simplify error handling too, but that's polish, so maybe a bit early.

I'm not sure how to approach this, without juggling with `void*`, it's all different types... well anyway, let me know if I can help with it somehow (but with more details what to do).

Bigfoot71:
> You are very welcome! It was fun to work with you, and I'm honored to have my code included to this awesome library. I'm going to use R3D in my indie project, so you'll probably hear from me here a lot in future... :)

Well that's good to hear, I mostly feel like I'm being annoying x)
But I'd be happy to hear how your project is going!

> I'm not sure how to approach this, without juggling with `void*`, it's all different types...

We can look at that later, it's not very critical right now.

But it's doable with something like this:
```c
size_t size = 0;
size += statesCount * sizeof(*((r3d_animtree_stm_t*)0)->nodeList);
size += edgesCount * sizeof(*((r3d_animtree_stm_t*)0)->edgeList);
/*...*/

char* block = RL_MALLOC(size);
if (!block) {
    /* handle failure */
}

r3d_animtree_stm_t* stm = anode->stm;
stm->nodeList = (void*)block; block += statesCount * sizeof(*stm->nodeList);
stm->edgeList = (void*)block; block += edgesCount * sizeof(*stm->edgeList);
/*...*/
```

---

Well, everything seems perfect as it stands, so let's merge!

Feel free to revisit what was done afterward, the next pre-release will take a little while, I'm going to take my time with this one.

And thank you again for this great contribution!

PR #224, opened 2026-03-03
Light toggle shadow check

grunthon:
`R3D_ToggleLight` had a redundant check for if the light was enabled instead of checking if it was using shadows when determining if the light's shadow should be flagged for update.


Bigfoot71:
Oh, well spotted, thanks!

PR #226, opened 2026-03-04
Fix anode_reset invalid return

grunthon:
`void` return type `anode_reset` was trying to return the result of other, also `void` return, functions.

I haven't had a chance to try or really look at the animation tree stuff. I just caught this in a compiler warning so this is a simple adjustment without looking for any possible underlying issues.

### DIFF
--- a/src/r3d_animation_tree.c
+++ b/src/r3d_animation_tree.c
@@ -472,12 +472,12 @@ static void anode_reset_stm_x(r3d_animtree_stm_x_t* node)
 static void anode_reset(R3D_AnimationTreeNode anode)
 {
     switch(anode.base->type) {
-    case R3D_ANIMTREE_ANIM:   return anode_reset_anim(anode.anim);
-    case R3D_ANIMTREE_BLEND2: return anode_reset_blend2(anode.bln2);
-    case R3D_ANIMTREE_ADD2:   return anode_reset_add2(anode.add2);
-    case R3D_ANIMTREE_SWITCH: return anode_reset_switch(anode.swch);
-    case R3D_ANIMTREE_STM:    return anode_reset_stm(anode.stm);
-    case R3D_ANIMTREE_STM_X:  return anode_reset_stm_x(anode.stmx);
+    case R3D_ANIMTREE_ANIM:   anode_reset_anim(anode.anim); break;
+    case R3D_ANIMTREE_BLEND2: anode_reset_blend2(anode.bln2); break;
+    case R3D_ANIMTREE_ADD2:   anode_reset_add2(anode.add2); break;
+    case R3D_ANIMTREE_SWITCH: anode_reset_switch(anode.swch); break;
+    case R3D_ANIMTREE_STM:    anode_reset_stm(anode.stm); break;
+    case R3D_ANIMTREE_STM_X:  anode_reset_stm_x(anode.stmx); break;
     default:
         R3D_TRACELOG(LOG_WARNING, "Failed to reset node: invalid type %d", anode.base->type);
         break;


misokobza:
Thank you @grunthon! There should be no underlaying issues, it was just copy/paste error.

grunthon:
@misokobza Again just looking at compiler warnings, but there are quite a few signed/unsigned comparisons in the animation tree functions. I might suggest keeping it simple and sticking with regular `int` types, or possibly `size_t` where it would be appropriate.

I'm sure if @Bigfoot71 has an opinion on this I'm sure he'll let you know or make the adjustments.

misokobza:
Damn, you are right... I didn't have warnings enabled in cmake, thought it's all on by default. My mistake, sorry...

Anyway, I was able to find another small bug thanks to 'unused variable' warning, so thank you again @grunthon! 👍

@Bigfoot71 pls let me know what to do with signed/unsigned mismatch, it should be now the only warning type in animation tree. I agree with @grunthon, currently it's a mess...

Bigfoot71:
> I might suggest keeping it simple and sticking with regular `int` types, or possibly `size_t` where it would be appropriate.

I think that anything like counters should indeed remain signed for simplicity, we have absolutely nothing to lose from doing so.

It seems to apply to all the unsigned types currently exposed in the API, so yeah, let's just switch everything to int.

---

> @Bigfoot71 pls let me know what to do with signed/unsigned mismatch, it should be now the only warning type in animation tree. I agree with @grunthon, currently it's a mess...

So yeah, honestly let's just switch everything to `int`. If there are complaints about casts with `stdlib`/`string`, you can use/cast to `size_t`, but only internally, there don't seem to be any other constraints, but I'll let you be the judge.

@misokobza Btw, I just noticed in `anode_reset_switch` that you're assigning `unsigned char` values from `unsigned int`, switching everything to `int` will definitely save us some hair ^^

https://github.com/Bigfoot71/r3d/blob/d3447facdb0e345e5c649d1cabf46560866146ef/src/r3d_animation_tree.c#L430-L433

That said, I wouldn't be that harsh, it's still good work!

misokobza:
> So yeah, honestly let's just switch everything to int. If there are complaints about casts with stdlib/string, you can use/cast to size_t, but only internally, there don't seem to be any other constraints, but I'll let you be the judge.

Sure, no problem.

> @misokobza Btw, I just noticed in anode_reset_switch that you're assigning unsigned char values from unsigned int, switching everything to int will definitely save us some hair ^^

It's all artefacts from my original "exposed API" version - there I have no dynamic allocation, and everything has dimensions under 255 (so almost all is unsigned char). It shows no issues when compiled with my Makefile with enabled warnings.

So again, sorry for that, I really thought warnings are enabled by default in r3d build, because during compilation it showed a lot of warnings for raylib stuff...

grunthon:
> I really thought warnings are enabled by default in r3d build, because during compilation it showed a lot of warnings for raylib stuff...

@misokobza Apologies if I gave the impression that you should be overly focused on compiler warnings or your code was a mess in some way. I only mentioned them because, as I said, I specifically hadn't actually reviewed the code and the issue addressed in this PR and then the sign mismatches are things I happened to noticed in the warnings.

The seriousness of compiler warnings can vary quite a lot depending on compiler, warning strictness, the specific case and sometimes even just personal opinion.

misokobza:
> @misokobza Apologies if I gave the impression that you should be overly focused on compiler warnings or your code was a mess in some way.

No problemo! I appreciate your help, plus I made a little mess with the types, not by purpose, but still... :)

PR #243, opened 2026-03-13
Update ray4laz_r3d version to 0.9 in README

GuvaCode:


Bigfoot71:
Wow that was fast! x)

GuvaCode:
> Wow that was fast! x)

Well, that's expected. I make edits as soon as they are added to the master branch.  Almost in real time. I worked with shaders. but I had to specify a larger size when compiling.

I worked with shaders. but I had to specify a larger size when compiling. And yes, that 's what the project was missing !!!

sp2
test1

PR #283, opened 2026-05-13
Use correct value for bloom soft threshold

grunthon:
The base threshold was incorrectly being used for the soft threshold.


Bigfoot71:
wow thanks for that!

PR #284, opened 2026-05-14
Set SSGI edge fade uniform as float

grunthon:
SSGI edge fading was not working due to the shader uniform being set as an integer instead of the correct float type.


Bigfoot71:
good catch, thanks !

PR #296, opened 2026-05-27
Fix ambient map initialization for C++ in R3D_ENVIRONMENT_BASE 

grunthon:
`R3D_ENVIRONMENT_BASE` was using C style initialization for `ambient.map` which was invalid for C++ projects. This adds a macro to do the assignment based on language.

I did not find any other cases where this was an issue and `ambient.map` could probably just be initialized with `{0}` to avoid adding the new macro but this keeps the type explicit if that is wanted.

### DIFF
--- a/include/r3d/r3d_environment.h
+++ b/include/r3d/r3d_environment.h
@@ -63,7 +63,7 @@
         .ambient = {                                    \
             .color = BLACK,                             \
             .energy = 1.0f,                             \
-            .map = (R3D_AmbientMap) {0},                \
+            .map = R3D_TYPEINIT(R3D_AmbientMap)         \
         },                                              \
         .ssao = {                                       \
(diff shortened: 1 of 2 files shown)


Bigfoot71:
Thank you for the correction, it's perfect!

~~Just a side note, `R3D_LITERAL(R3D_AmbientMap){0}` could have been used, but this is more readable like this.~~

Oh yeah, sorry I'm tired, there's no need to write the type, I'll fix that.

Fixed: 3ec6b193d544a8e6d28037e4ebaeba24e04bc90a be8e1d5c530a043ec1c5cf6c6172f3c12601529e

grunthon:
> there's no need to write the type, I'll fix that.

Yeah that's the most efficient. I just quickly did this while sticking to the existing style. I maybe should have made an issue instead.

PR #304, opened 2026-06-18
Light and probe API redesign

Bigfoot71:
## Overview

As we approach the final release, the public API will undergo a series of
breaking changes to ensure consistency, clarity and long-term stability.

This PR is the first step of that process, focusing on the lighting and
probe modules.

### Naming

- `R3D_IsLightExist` -> `R3D_IsLightValid`
- `R3D_IsLightActive` -> `R3D_IsLightEnabled`
- `R3D_SetLightActive(id, bool)` -> `R3D_EnableLight(id)` / `R3D_DisableLight(id)`
- `R3D_LightLookAt` -> `R3D_SetLightTarget`
- `R3D_GetLightAttenuation` / `R3D_SetLightAttenuation` -> `R3D_GetLightFalloff` / `R3D_SetLightFalloff`
- `R3D_GetShadowUpdateFrequency` / `R3D_SetShadowUpdateFrequency` -> `R3D_GetShadowUpdateInterval` / `R3D_SetShadowUpdateInterval`
- `R3D_DrawLightShape` -> `R3D_DrawLightDebug`
- `R3D_IsProbeExist` -> `R3D_IsProbeValid`
- `R3D_IsProbeActive` -> `R3D_IsProbeEnabled`
- `R3D_SetProbeActive(id, bool)` -> `R3D_EnableProbe(id)` / `R3D_DisableProbe(id)`

### Spot light cone angles

`SetLightInnerCutOff` and `SetLightOuterCutOff` have been merged into a single
call to lighten the API and prevent invalid states where `inner > outer`:

```c
void R3D_SetLightAngle(R3D_Light id, float inner, float outer); // degrees
void R3D_GetLightAngle(R3D_Light id, float* inner, float* outer);
```

### Shadow update interval

Now expressed as a `float` in seconds instead of an `int` in milliseconds,
consistent with `GetFrameTime()`.

### Distance attenuation model

The attenuation formula has been replaced by a power curve:

```glsl
// Before
float atten = (1.0 - clamp(Ldist / light.range, 0.0, 1.0)) * light.attenuation;

// After
float atten = pow(1.0 - clamp(Ldist / light.range, 0.0, 1.0), light.falloff);
```

The falloff exponent is now stored and exposed directly without inversion.
A value of `1.0` is linear, `2.0` is quadratic. Values at or below `0.0`
are clamped to `1.0`.

**Existing attenuation values are not compatible and must be reviewed.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ int main(void)
     // Setup lighting
     R3D_Light light = R3D_CreateLight(R3D_LIGHT_DIR);
     R3D_SetLightDirection(light, (Vector3){-1, -1, -1});
-    R3D_SetLightActive(light, true);
+    R3D_EnableLight(light);
 
     // Camera setup
     Camera3D camera = {
--- a/examples/animation.c
+++ b/examples/animation.c
@@ -55,7 +55,7 @@ int main(void)
     // Setup lights with shadows
     R3D_Light light = R3D_CreateLight(R3D_LIGHT_DIR);
     R3D_SetLightDirection(light, (Vector3){-1.0f, -1.0f, -1.0f});
-    R3D_SetLightActive(light, true);
+    R3D_EnableLight(light);
     R3D_SetLightRange(light, 10.0f);
     R3D_EnableShadow(light);
 
--- a/examples/animtree.c
+++ b/examples/animtree.c
@@ -119,7 +119,7 @@ int main(void)
     // Setup lights with shadows
     R3D_Light light = R3D_CreateLight(R3D_LIGHT_DIR);
     R3D_SetLightDirection(light, (Vector3){-1.0f, -1.0f, -1.0f});
-    R3D_SetLightActive(light, true);
+    R3D_EnableLight(light);
     R3D_SetLightRange(light, 10.0f);
     R3D_EnableShadow(light);
 
--- a/examples/basic.c
+++ b/examples/basic.c
@@ -1,3 +1,4 @@
+#include "r3d/r3d_lighting.h"
 #include <r3d/r3d.h>
 #include <raymath.h>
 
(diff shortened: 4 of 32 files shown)


grunthon:
> R3D_DrawLightShape -> R3D_DrawLightGizmo

`R3D_DrawLightShape` looks to have been changed to `R3D_DrawLightDebug` and not `R3D_DrawLightGizmo` as listed here.

> The falloff exponent is now stored and exposed directly without inversion.
> A value of 1.0 is linear, 2.0 is quadratic. Values at or below 0.0
> are clamped to 1.0.

I was curious about setting a falloff of <= 0 back up to 1 but looking at `R3D_SetLightFalloff`  it is actually being clamped between 0 and 1.
[`light->falloff = Clamp(falloff, 0.0f, 1.0f);`](https://github.com/Bigfoot71/r3d/blob/a6ff364c4f31368fe9cf1feb1eaa2e9569add552/src/r3d_lighting.c#L225)

Setting <= 0 to 1, or possibly a preset epsilon, would seem to be what is intended.
`light->falloff = (falloff <= 0.0f) ? 1.0f : falloff;`

Though I'm not sure if this is even really needed or particularly intuitive. In general the library already leaves it to the user to maintain reasonable values. Even with the functionality documented things like:
```
R3D_SetLightFalloff(l, 0.0f);
float f = R3D_GetLightFalloff(l);
// f is now 1.0, or epsilon
```
could result in trickier problems to track down than just allowing "incorrect" values.

Either way, it is currently not functioning as intended.

Bigfoot71:
Thanks for the review!

For the Gizmo, I just mixed up my notes, I first used that name, then renamed it to something less "editorish".

For falloff, yeah that's clearly a mistake, it indeed should've been `(falloff <= 0.0f) ? 1.0f : falloff`
I should've caught that during my review...

The intent was just to avoid Inf/NaN cases like `pow(0, -1)` or even `pow(0, 0)`, which I think may end up as `exp(0 * log(0))`, because on my side that also gives NaN instead of 1.0

So I was thinking it might be better to clamp it and document it, since either way the math is kind of undefined in those cases.

I'm going to update the comment, and I've updated the setter: https://github.com/Bigfoot71/r3d/commit/4eee4095a3dbdec744f7d2f830c8f008801e3ee7

---

Edit: I might go over it again later, if you have any suggestions for breaking changes, don't hesitate, now's the time or never!

Also next time, if there is an API change, I'll leave the PRs pending for a bit and do at least two review passes just in case...

PR #306, opened 2026-06-20
Volumetric Fog

Bigfoot71:
## Overview

This PR adds a physically-based volumetric fog effect to R3D,
computed at half resolution via multiple passes for performance.

The effect simulates how light scatters through fog or haze, supporting all
three light types (directional, spot, omni) including shadow-casting lights.
Each light contributes visible rays and glow through the volume, with an
anisotropy parameter to control whether light scatters evenly or favors a
particular direction.

The pipeline consists of three passes:

- **Transmittance**: analytical extinction applied directly to the scene via
  hardware blending, with optional emission color to tint the fog volume
  independently of lighting.
- **Radiance**: per-light ray-marching at half resolution, accumulated
  additively into a shared buffer using scissor-culled draw calls.
- **Composition**: depth-aware bilateral upsampling with rotated-grid filtering
  to reconstruct the radiance buffer at full resolution without too much visible
  half-resolution artifacts.

**Current limitation:** volumetric fog is applied to opaque geometry and the
skybox only. Transparent objects are not yet supported, as doing so requires
additional pipeline considerations to handle correctly and efficiently. As a
temporary workaround, setting `emissionEnergy` and `absortionDensity` to zero
alongside an atmospheric fog can provide an acceptable result in scenes where
transparent object interaction is a concern.

The atmospheric fog pipeline has also been revised; skybox-specific fog logic
has been removed from the background pass, and atmospheric fog for both opaque
geometry and the background is now applied in a single pass after the background
and before transparent objects, like how volumetric fog is composed.

## Changes

- Added `R3D_VolumetricFog` struct to `R3D_Environment` in the public API.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,6 +353,9 @@ embed_shaders(${PROJECT_NAME}
     "${R3D_ROOT_PATH}/shaders/deferred/lighting.frag"
     "${R3D_ROOT_PATH}/shaders/deferred/compose.frag"
     "${R3D_ROOT_PATH}/shaders/deferred/fog.frag"
+    "${R3D_ROOT_PATH}/shaders/deferred/vfog_transmittance.frag"
+    "${R3D_ROOT_PATH}/shaders/deferred/vfog_radiance.frag"
+    "${R3D_ROOT_PATH}/shaders/deferred/vfog_compose.frag"
     # Post
     "${R3D_ROOT_PATH}/shaders/post/dof.frag"
     "${R3D_ROOT_PATH}/shaders/post/bloom.frag"
--- a/include/r3d/r3d_environment.h
+++ b/include/r3d/r3d_environment.h
@@ -104,11 +104,23 @@
         },                                              \
         .fog = {                                        \
             .mode = R3D_FOG_DISABLED,                   \
-            .color = {255, 255, 255, 255},              \
+            .color = WHITE,                             \
             .start = 1.0f,                              \
             .end = 50.0f,                               \
             .density = 0.05f,                           \
             .skyAffect = 0.5f,                          \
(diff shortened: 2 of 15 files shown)


Bigfoot71:
Some very quick examples:

[ex1](https://github.com/user-attachments/assets/75e648ef-1a3d-4c7c-a3fa-8c6a7dc20678)

[ex2](https://github.com/user-attachments/assets/6ef76631-ed62-48f9-afd0-24c1b4e2a915)

grunthon:
I see there was a commit to clamp ray marching with light range but there does seem to be an issue with light distance and range.

Here is the sponza example with the one of the lights far enough away to be completely out of range.
`R3D_SetLightPosition(lights[i], (Vector3) { i ? -100.0f : 10.0f, 20.0f, 0.0f });`

I haven't look any further into it but increasing the light's range does seem to correct it so it would seem to be out of range lights being incorrectly used.

https://github.com/user-attachments/assets/534861af-f542-47e3-8724-82e1c4f0b731

Bigfoot71:
> I haven't look any further into it but increasing the light's range does seem to correct it so it would seem to be out of range lights being incorrectly used.

Thanks for reporting it, I just noticed the issue myself in my game too!

The problem actually came from the scissor state remaining incorrect between passes, it's now fixed.

There was also an issue with the range clipping, I went way too fast when I had the idea, that's been fixed too.
It wasn't the cause of this problem, it could have just clipped the fog incorrectly at distance in certain cases.

grunthon:
> The problem actually came from the scissor state remaining incorrect between passes, it's now fixed.

Ah. That does make sense. The way it accumulated in specific areas did seem odd for just a bad light value in the shaders.

PR #320, opened 2026-08-06
Replace ID-based light/probe management with stateless API

Bigfoot71:
Part of the work to prepare the official release.

Removes the ID-based system for managing lights (and next probes) in favor of an immediate-mode / stateless style. Lights are now pushed every frame between `R3D_Begin()`/`R3D_End()` instead of being created/updated/destroyed via global IDs.

The old ID-based system operated globally, which would have made it awkward to later scope lights/probes to independent scene contexts. This immediate-mode approach removes that coupling and paves the way for an optional multi-scene system, while greatly simplifying the API.

Follow-up work planned:
- Add finer control over when shadows get updated (still an open question).
- Decouple probes into separate GI and reflection systems.
- Reduce internal per-frame light copies.

**Trade-offs:**
- No more dirty-flag-based shadow update skipping, in practice negligible cost.
- Multiple per-frame light data copies, further optimization possible later.

_**Note:** This PR isn't intended to be merged quickly, any feedback is appreciated. I'll likely make another pre-release in the meantime with the latest added features and API changes, excluding these changes._


Bigfoot71:
As a quick example, lights are now managed via structures that can be fully populated manually:
```c
typedef struct R3D_Light {
    Vector3 position;       ///< Light position (spot/omni)
    Vector3 direction;      ///< Light direction (spot/dir)
    Color color;            ///< Light color
    float energy;           ///< Light intensity/brightness multiplier
    float specular;         ///< Specular reflection intensity multiplier
    float range;            ///< Maximum distance (spot/omni)
    float falloff;          ///< Distance falloff factor (spot/omni)
    float innerCutOff;      ///< Spot light inner cutoff angle (degrees)
    float outerCutOff;      ///< Spot light outer cutoff angle (degrees)
    float fogEnergy;        ///< Volumetric fog energy multiplier
    R3D_LightType type;     ///< Light type (directional/spot/omni)
} R3D_Light;
```

Here is a quick overview of how to use them currently:
```c
R3D_Light sun = R3D_CreateDirLight((Vector3){-1, -1, -0.5f}, WHITE, 1.0f);
R3D_Light spot = R3D_CreateSpotLight((Vector3){0, 10, 5}, (Vector3){0, -1, -0.5f}, 20.0f, WHITE, 5.0f);

R3D_ShadowMap sunShadow = R3D_LoadShadowMap(R3D_LIGHT_DIR);

// Render loop

R3D_Begin(camera);

    R3D_DrawModel(model, (Vector3){0}, 1.0f);

    R3D_PushLightEx(sun, sunShadow, updateShadow);   // with shadows
    R3D_PushLight(spot);                             // without shadows

R3D_End();

// End of the program

R3D_UnloadShadowMap(sunShadow);
```

Bigfoot71:
Well, I've finished the "frontend" work, there are still some things that could be done to improve it, but I prefer doing it separately in the end.

I'd like this to become a stable base before we build more on top of it _(missing helpers can always be added later, that's not a concern)_.

If you have thoughts on **naming, parameters, or behavior of what's here**, now's the time, please comment below.

_@grunthon I'm pinging you in case you have an opinion on this_

---

So, now you can create/fill your `R3D_Light`/`R3D_Probe`/`R3D_ShadowMap` structures and you should `Push` it every frame between `R3D_Begin()`/`R3D_End()`.

**Light creation helpers:**
```c
R3D_Light R3D_CreateDirLight(Vector3 dir, Color color, float energy);
R3D_Light R3D_CreateSpotLight(Vector3 pos, Vector3 dir, float range, Color color, float energy);
R3D_Light R3D_CreateOmniLight(Vector3 pos, float range, Color color, float energy);
```

**New loading functions:**
```c
R3D_ShadowMap R3D_LoadShadowMap(R3D_LightType type);
void R3D_UnloadShadowMap(R3D_ShadowMap map);

// Note: a probe is now either illumination OR reflection, not both.
// The illumination method for probes may change later, hence the split.
R3D_Probe R3D_LoadProbe(R3D_ProbeType type, bool interior, bool shadow);
void R3D_UnloadProbe(R3D_Probe probe);
```

**Per-frame submission:**
```c
void R3D_PushLight(R3D_Light light);
void R3D_PushLightEx(R3D_Light light, R3D_ShadowMap map, bool updateShadow);
void R3D_PushLights(const R3D_Light* lights, int count);
void R3D_PushProbe(R3D_Probe probe, bool updateProbe);
```
`updateShadow`/`updateProbe` let you skip re-rendering a shadow/capture on frames where nothing changed (e.g. static lights), but the first use always renders regardless, so passing `false` from the start is always safe.

Full doc comments are in the diff.

Bigfoot71:
Note that, currently, setting `update` to `true` does not guarantee an update if the probe (or shadow-casting light) is eliminated by frustum culling. ~~I need to give this case some further thought as well.~~

Fixed:
- [1d2e01b](https://github.com/Bigfoot71/r3d/pull/320/commits/1d2e01be6bb9ea575cf2d4311966f79c9f3aa204)
- [158be35](https://github.com/Bigfoot71/r3d/pull/320/commits/158be359b7624c8fe3ea31543aaea453a9ba4ddd)
- [aa402ea](https://github.com/Bigfoot71/r3d/pull/320/commits/aa402eac6578478f5b861fc2100780d15edc455d)

grunthon:
I haven't really had time to really look at this but appreciate the ping.

One thing I do wonder about is the shadow map being separate from the light. I know having the shadow map stored or referenced in `R3D_Light` could cause real problems since it's in a user owned object, but it does put more to manage on the user. It also leads to needing the separate PushLight functions and seems to make `R3D_PushLights` less useful.

As far as naming, the Load/Unload functions for shadow maps (and probably probes) seem maybe inconsistent as nothing is actually being loaded in the manner of other Load functions like `R3D_LoadAlbedoMap`, etc. Perhaps something like Create/Destroy would be more appropriate?

Just quick thoughts and I should have more time to think about it later this weekend (particularly the lights/shadows) and maybe have more useful feedback.

Bigfoot71:
> One thing I do wonder about is the shadow map being separate from the light.

Yeah, there is definitely an asymmetry in the `Push` functions. I think we could already reduce some of that friction by removing `PushLights`. I originally had it in mind for batches of dynamic lights that generally don't cast shadows anyway, but it only saves us one loop.

As for keeping lights and shadows separate, after thinking about it more, I feel that combining them could actually make the lifecycle more confusing.

Keeping them separate at least gives full flexibility over their usage and enforces a clear separation between light and shadow allocation management. But I'm definitely open to better ideas.

> As far as naming, the Load/Unload functions for shadow maps (and probably probes) seem maybe inconsistent

I agree that it feels a bit awkward too, but I'm trying to follow raylib's naming conventions, which also uses `Load`/`Unload` for pretty much everything, including things like `LoadRenderTexture()`.

Bigfoot71:
I just released pre-release 0.11.0 without these API changes, giving us a specific tag for the latest features and bug fixes while we focus on the internal and API work for the next release.

I'll merge this PR now so I can move on with the rest of my TODO. I'll reference any new PRs related to the light or probe APIs here if I make further changes, so we can keep track of everything.

Any issues or comments here are still very welcome!

PR #321, opened 2026-08-07
Fix: R3D_RaycastModel check local ray instead of global

DuyhaBeitz:
In R3D_RaycastModel global ray was used to check collision with local space mesh.aabb

Before:
before
After:
after


Bigfoot71:
Thank you!

PR #327, opened 2026-08-13
fix `R3D_AnimationTree`: corrected xFade time threshold value

misokobza:
xFade time threshold value in animation node must scale with animation speed to work correctly with state machine


Bigfoot71:
Thanks for the fix!

PR #332, opened 2026-08-19
fix `R3D_AnimationTree`: added missing reset of last root position in animation node

misokobza:
Missing reset manifested as "root position jumping" when switching root-sensitive animations.


Bigfoot71:
Thank you!

